PR #52, opened 2018-08-03
Add endpoints to create and get update objects.

houcros:


src/main/resources/db/migration/V5__new_table_updates.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE `updates` (
+  `uuid` CHAR(36) NOT NULL,
+  `namespace` CHAR(200) NOT NULL,
+  `name` VARCHAR(200) NOT NULL,
+  `description` VARCHAR(600) NOT NULL,
+  `created_at` DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `updated_at` DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+
```
koshelev: We need an optional external ID here

houcros: Alright, I added an `externalID` with a unique constraint on `(ns, externalId)`.

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
```diff
@@ -37,19 +36,17 @@ object DataType {
 
   final case class Update(
                            id: UpdateId,
+                           externalId: String,
```
koshelev: please make it `Option[String]`

houcros: Done :)

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
```diff
@@ -35,6 +34,21 @@ object DataType {
...
+                           externalId: Option[String],
+                           namespace: Namespace,
+                           name: String,
+                           description: Option[String],
+                           createdAt: Instant,
+                           updatedAt: Instant
+                         )
+
+  final case class CreateUpdate(externalId: Option[String], name: String, description: Option[String]) {
+    def mkUpdate(ns: Namespace): Update =
```
kmanikandan-kmk: Shall we change this to makeUpdate() or createUpdate() or something like that? mkUpdate() does not seem to be perfect.

houcros: Can do :) But in that case I would propose to rename `mkCampaign` and `mkCampaignMetadata` as well to be consistent.
Personally I'd go for `build(Update|Campaign|CampaignMetadata>`.

koshelev: `mkSomething` is used across the whole codebase, so it is ok

kmanikandan-kmk: I would also go for build(Update|Campaign|CampaignMetadata). I will leave it to Jerry/Simao to decide.

src/main/scala/com/advancedtelematic/campaigner/db/Schema.scala
```diff
@@ -95,4 +98,30 @@ object Schema {
...
+
+  class UpdatesTable(tag: Tag) extends Table[Update](tag, "updates"){
+    def uuid = column[UpdateId]("uuid", O.PrimaryKey)
+    def updateId = column[String]("update_id")
+    def updateSourceType = column[UpdateType]("update_source_type")
+    def namespace = column[Namespace]("namespace")
+    def name = column[String]("name")
+    def description = column[Option[String]]("description")
+    def createdAt = column[Instant]("created_at")
+    def updatedAt = column[Instant]("updated_at")
```
simao: we generally don't need timestamps at the app level, I don't know why `Campaign` uses them. The db should have them so we can audit, but they are not needed on the app level.

koshelev: As we are going to publish our APIs (god help us), it is probably not a bad idea to have those for a campaign. It is not clear what "updated" means though...

PR #53, opened 2018-08-07
[OTA-268] Add campaign metadata types

jerrytrieu:


src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
```diff
@@ -20,7 +20,9 @@ object DataType {
 
   object MetadataType extends Enumeration {
     type MetadataType = Value
-    val install = Value
+    val description = Value
+    val estimatedInstallationDuration = Value
+    val estimatedPreparationDuration = Value
   }
```
simao: This can be all in one line `description, estimatedInstallationDuration, estimatedPreparationDuration = Value`

The migration will fail if there are already metadata for some campaigns with `install`.

 I think the capitalization is a bit weird maybe should be `EstimatedInstallationDuration`, or `ESTIMATED_INSTALL_DURATION` ?

houcros: Capitalization for enums seems to follow that style (at least in campaigner).

koshelev: +1 for capitalization

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
```diff
@@ -20,7 +20,7 @@ object DataType {
 
   object MetadataType extends Enumeration {
     type MetadataType = Value
-    val install = Value
+    val description, estimatedInstallationDuration, estimatedPreparationDuration = Value
```
houcros: How do we feel about this capitalization after all?

koshelev: I do like capitalization for enums. But do not forget to adjust the DB schema too

houcros: Oh I see now about my previous PR. Alright, it's capitalized now.

PR #54, opened 2018-08-07
Wait for accept status before setting MTU in director

simao:


jerrytrieu:
Rebased and resolved conflicts.

build.sbt
```diff
@@ -25,7 +25,7 @@ resolvers += "ATS Snapshots" at "http://nexus.advancedtelematic.com:8081/content
 libraryDependencies ++= {
   val akkaV = "2.5.13"
   val akkaHttpV = "10.1.1"
-  val libatsV = "0.1.2-20-ge2096d9"
+  val libatsV = "0.1.2-22-g5f08951"
```
houcros: libats is at 27 already, maybe a good chance to bump it up?

simao:
@jerrytrieu  Added autoAccept attribute to create api  

Bumped libats

src/main/resources/db/migration/V5__auto_accept_campaigns.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE `campaigns` ADD COLUMN
+`auto_accept` BOOL NOT NULL DEFAULT TRUE
+;
+
+ALTER TABLE `device_updates` MODIFY COLUMN
+`status` ENUM('scheduled', 'accepted', 'successful', 'cancelled', 'failed') NOT NULL
```
kmanikandan-kmk: I think, this will lead to an API endpoint of createCampaign(name, desc, ... , autoAccept = TRUE).

IMO, as API description, it is good to have an endpoint like createCampaign(name, desc, ... , approvalNeeded = FALSE) so that consumer of this API (for example: webapp) can call with approvalNeeded = TRUE/FALSE accordingly. approvalNeeded will be FALSE by default.

simao: changed

jerrytrieu:
I will remove last commit and rebase and merge.

PR #56, opened 2018-08-09
Feat/ota 385/updates resource

koshelev:


src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
```diff
@@ -134,7 +134,7 @@ object DataType {
 
   object UpdateType extends Enumeration {
     type UpdateType = Value
-    val EXTERNAL, MULTI_TARGET = Value
+    val external, multi_target = Value
```
houcros: `multitarget`?   `multiTarget`?

PR #58, opened 2018-08-16
Migration to extract and create updates from existing campaigns.

houcros:


src/main/scala/com/advancedtelematic/campaigner/db/ExtractUpdatesFromCampaignsAndInsert.scala
```diff
@@ -0,0 +1,49 @@
...
+    db.run(campaigns.result)
+      .map { cs => cs.map(updateFromCampaign) }
+      .map { us => us.groupBy(_.source).map(_._2.head) } // filter UpdateSource duplicates
+      .map { us => us.groupBy(u => (u.namespace, u.source.id)).map(_._2.head) } // filter unique key (namespace, updateId) duplicates (shouldn't be necessary after the previous filter?)
+      .flatMap { us => db.run(updates ++= us) }
+
+  def run() = {
+    val campaignIdUpdateIdPairs = for {
+      (c, u) <- campaigns join updates on (_.update.mappedTo[String] === _.updateId)
+    } yield (c.id, u.uuid)
```
simao: we try to avoid infix notation https://docs.scala-lang.org/style/method-invocation.html 

In this case would be ok, but in general is best to avoid it.

Also this would be easier to read with:

```
    val campaignIdUpdateIdPairs = campaigns.join.updates.on(_.update.mappedTo[String] === _.updateId).map { case (c, u) => (c.id, u.uuid) }
```

src/main/scala/com/advancedtelematic/campaigner/db/ExtractUpdatesFromCampaignsAndInsert.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  private val _log = LoggerFactory.getLogger(this.getClass)
+
+  private[db] def updateFromCampaign(c: Campaign): Update = {
+    val u = Update(UpdateId.generate(), UpdateSource(c.updateId.uuid.toString, UpdateType.multi_target), c.namespace, c.name, None, Instant.now, Instant.now)
+    _log.info("Extracted update {} from campaign {}", u, c, "")
+    u
+  }
+
+  private[db] def createOneUpdateRecordForEachCampaign(): Future[Option[Int]] =
```
simao: If I am reading this right, this would be the same as:

```
    db.run(campaigns.distinctOn(_.update).result)
      .map { cs => cs.map(updateFromCampaign) }
      .flatMap { us => db.run(updates ++= us) }
```
?

Also, this works because this table is relatively small, but in the future we should write migrations using `db.stream` + async writes to the database so we don't bring the whole table into memory and write it back to mysql. For example see: https://github.com/advancedtelematic/ota-tuf/blob/e54bd1770fa82594887e19ca1222dc0b85d0be0b/libtuf-server/src/main/scala/com/advancedtelematic/libtuf_server/db/SignatureMethodMigration.scala#L33

src/main/scala/com/advancedtelematic/campaigner/db/ExtractUpdatesFromCampaignsAndInsert.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  def run() = {
+    val campaignIdUpdateIdPairs = for {
+      (c, u) <- campaigns join updates on (_.update.mappedTo[String] === _.updateId)
+    } yield (c.id, u.uuid)
+
+    createOneUpdateRecordForEachCampaign()
+      .flatMap { _ => db.run(campaignIdUpdateIdPairs.result) }
+      .flatMap { pairs => Future.traverse(pairs)(pair => db.run(campaigns.filter(_.id === pair._1).map(_.update).update(pair._2))) }
+  }
```
simao: We should find a way for all this to run in a single migration.

src/main/scala/com/advancedtelematic/campaigner/db/ExtractUpdatesFromCampaignsAndInsert.scala
```diff
@@ -0,0 +1,49 @@
...
+  private val _log = LoggerFactory.getLogger(this.getClass)
+
+  private[db] def updateFromCampaign(c: Campaign): Update = {
+    val u = Update(UpdateId.generate(), UpdateSource(c.updateId.uuid.toString, UpdateType.multi_target), c.namespace, c.name, None, Instant.now, Instant.now)
+    _log.info("Extracted update {} from campaign {}", u, c, "")
+    u
+  }
+
+  private[db] def createOneUpdateRecordForEachCampaign(): Future[Option[Int]] =
+    db.run(campaigns.result)
```
simao: we should delay db.run as long as possible, all these ops can be done on DBIO level.

src/main/scala/com/advancedtelematic/campaigner/db/ExtractUpdatesFromCampaignsAndInsert.scala
```diff
@@ -0,0 +1,49 @@
...
+      .map { cs => cs.map(updateFromCampaign) }
+      .map { us => us.groupBy(_.source).map(_._2.head) } // filter UpdateSource duplicates
+      .map { us => us.groupBy(u => (u.namespace, u.source.id)).map(_._2.head) } // filter unique key (namespace, updateId) duplicates (shouldn't be necessary after the previous filter?)
+      .flatMap { us => db.run(updates ++= us) }
+
+  def run() = {
```
simao: So if I am reading this correctly, run could be:

```scala
  def run() = db.run {
    campaigns.distinctOn(_.update).result
      .map { cs => cs.map(c => c.id -> updateFromCampaign(c)).toMap }
      .flatMap { us => (updates ++= us.values).map(_ => us) }
      .flatMap { us =>
        DBIO.sequence(
          us.foldLeft(List.empty[DBIO[Int]]) { case (acc, (campaignId, update)) =>
            campaigns.filter(_.id === campaignId).map(_.update).update(update.uuid) :: acc
          }
        )
      }
      .transactionally
      .map(_ => Done)
  }
```

src/main/scala/com/advancedtelematic/campaigner/db/ExtractUpdatesFromCampaignsAndInsert.scala
houcros: Thanks, I'll keep it in mind.

src/main/scala/com/advancedtelematic/campaigner/db/ExtractUpdatesFromCampaignsAndInsert.scala
houcros: Awesome! I spent so many hours trying to come up with a "all in one go" query like that :D Indeed it looks like what we need, I tested it.
I just changed the first line with a workaround because there seems to be a bug with `distinctOn` and it was giving me duplicates, have a look at the comment in the code.

koshelev: that is exactly why I don't like Slick

simao: That is not a very good reason not to like slick... you could just write this in sql as well. https://github.com/advancedtelematic/ota-tuf/commit/6fd15d0#diff-cb2654b1aa07bb1bff4cf75ba0f3e614R23 

But anyway, yeah this is shitty. You can leave the workaround there because data is at the moment very small but this could easily blow up, and you would not be able to do something like this with db.stream.

src/main/scala/com/advancedtelematic/campaigner/db/ExtractUpdatesFromCampaignsAndInsert.scala
```diff
@@ -0,0 +1,50 @@
...
+  private val _log = LoggerFactory.getLogger(this.getClass)
+
+  private[db] def updateFromCampaign(c: Campaign): Update = {
+    val u = Update(UpdateId.generate(), UpdateSource(c.updateId.uuid.toString, UpdateType.multi_target), c.namespace, c.name, None, Instant.now, Instant.now)
+    _log.info("Extracted update {} from campaign {}", u.uuid, c.id, "")
+    u
+  }
+
+  def run(): Future[Done.type] = db.run {
+    campaigns.joinLeft(updates).on(_.update === _.uuid).filter(_._2.isEmpty).map(_._1)
```
simao: This one is needed to make this migration idempotent.

src/main/scala/com/advancedtelematic/campaigner/db/ExtractUpdatesFromCampaignsAndInsert.scala
```diff
@@ -0,0 +1,50 @@
...
+      .result
+      .map { cs => cs.groupBy(_.updateId) }
```
simao: we cannot use .head here, otherwise not all campaigns are updated down below, only the first one.

jerrytrieu:
Included in #62

PR #59, opened 2018-08-16
Use external resolver to resolve updates for device group

simao:


src/main/scala/com/advancedtelematic/campaigner/http/UpdateResource.scala
```diff
@@ -48,8 +52,18 @@ class UpdateResource(extractNamespace: Directive1[Namespace])
     HypermediaResource(links, update)
   }
 
+  val groupUpdateResolver = new GroupUpdateResolver(deviceRegistry, resolver)
+
+  import com.advancedtelematic.libats.http.UUIDKeyAkka._
+
   val route: Route =
     extractNamespace { ns =>
+      pathPrefix("group_updates" / GroupId.Path) { groupId =>
```
simao: @jerrytrieu what about `group_updates/<group_id>` ?

koshelev: Idiomatic way in this case would be  `/updates?groupId=<group_id>`

simao: That is what I hard at first...

jerrytrieu: That looks too simple for a complex query. Should be explicit `updates_for` or `applicable_updates`.

koshelev: @jerrytrieu `/updates?groupId=<group_id>` says exactly this.
IMO the best way would be `/campaigns?groupId=<group_id>` and returning an update as part of a campaign aggregate. The reason is there is no relation between an update and an update outside of context of a campaign

simao: *had

I don't think the api url should say how hard it is to get the resources, or any other detail of the implementation for that matter. 

@koshelev That would complicate getting the results but I guess would be possible... you mean something like `/campaigns/<campaign_id>?groupId=<group_id>` ? Since we already had get on `updates` I thought just returning that result would be fine.

koshelev: @simao If I understand it correctly, the endpoint should return all updates ever applied to a group. As every campaign is the link between an update and a group, we don't need a `campaign_id`. Or am I missing something?

simao: ah I see what you mean, yes ok..

koshelev: Ok, so I misunderstood what the endpoint is for. It should return **possible** updates for a group, not already scheduled updates. So `/updates?groupId=<group_id>` should do.

simao: So... should I revert to what I had at first or?

koshelev: AFAIU you had `/updates?groupId=<group_id>`, right? In this case it would be great if you revert.

simao: @jerrytrieu  ?

jerrytrieu: I don't see `GET /updates?groupId=id` and think `returns all updates that apply to all vehicles in a device group by querying an external resolver`. The information for that query doesn't exist under `/updates` resource. But I agree about not thinking about implementation in the API uri, so okay.

simao: hm I see your point.. we could add `applicable_updates` then we do `applicable_updates?groupId=<group_id>` this would expose the  applicable_updates resource which kind of makes sense.

koshelev: We can. But it feels like documenting endpoints with URIs. And as I mentioned, it is unlikely we will need to request applied updates outside of campaign's context.
What about a compromise: `GET /updates?groupId=<id>&applicable` and `GET /updates?groupId=<id>&applied`?

simao:
rebased and reverted back to `updates?groupId=<group_id>`

src/main/scala/com/advancedtelematic/campaigner/http/GroupUpdateResolver.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class GroupUpdateResolver(deviceRegistry: DeviceRegistryClient, resolver: Resolver)
+                         (implicit db: Database, ec: ExecutionContext) extends UpdateSupport  {
+
+  def groupUpdates(ns: Namespace, groupId: GroupId, maxRequests: Int = 50): Future[Seq[Update]] = {
+    def fetchDevices(offset: Long, limit: Long, requests: Int = 1): Future[Seq[DeviceId]] = {
+      deviceRegistry.devicesInGroup(ns, groupId, offset, limit).flatMap { devicesInBatch =>
+        if (devicesInBatch.isEmpty || requests >= maxRequests)
```
koshelev: Not sure it is correct behaviour. IMO if we have not managed to fetch all devices it should fail.

simao:
Fixed that and rebased

PR #60, opened 2018-08-20
Fix naming V7__auto_accept_campaigns.sql

jerrytrieu:
See 5c8adb5981dd975d4ec1c65b419b9614e43f3e7f


simao:
This will break flyway if migrations were already executed...

jerrytrieu:
This is to unbreak already executed flyway. I renamed the wrong file when rebasing previous PR. :(

simao:
ok but then I think you need a repair anyway

PR #62, opened 2018-08-22
New FK from campaigns to updates & migration to create updates from campaigns

houcros:
So we discussed that there should be a restriction so that each campaign must have an update object associated.
I created a new FK from campaigns to updates to enforce this. Then I had to refactor all the tests so that each time we create a campaign it has an existing update beforehand.
There is some code duplication which I still wasn't able to extract after much work, particularly the method `buildCampaignWithUpdate` is spread in a few classes. I wanted to extract it into a trait and mix it in but I didn't manage. So there is room for improvement.


src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
```diff
@@ -213,7 +216,11 @@ protected class CampaignRepository()(implicit db: Database, ec: ExecutionContext
 
   def persist(campaign: Campaign, groups: Set[GroupId], metadata: Seq[CampaignMetadata]): Future[CampaignId] = db.run {
     val f = for {
-      _ <- (Schema.campaigns += campaign).handleIntegrityErrors(Errors.ConflictingCampaign)
+      _ <- (Schema.campaigns += campaign).asTry.flatMap {
```
houcros: Also, is there a better way to do this? It feels hacky, but I wanted to differentiate and throw a different error when the cause is that there is no valid FK from campaigns to updates.

src/main/scala/com/advancedtelematic/campaigner/db/ExtractUpdatesFromCampaignsAndInsert.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  private val _log = LoggerFactory.getLogger(this.getClass)
+
+  private[db] def updateFromCampaign(c: Campaign): Update = {
+    val u = Update(UpdateId.generate(), UpdateSource(ExternalUpdateId(c.updateId.uuid.toString), UpdateType.multi_target), c.namespace, c.name, None, Instant.now, Instant.now)
+    _log.info("Extracted update {} from campaign {}", u.uuid, c.id, "")
+    u
+  }
+
+  def run(): Future[Done.type] = db.run {
```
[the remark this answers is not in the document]
houcros: If there is no argument agains reusing `updateId` then I say yes, let's do it like this, much easier to read.
Isn't it `select c.update_id, c.update_id [...]` ?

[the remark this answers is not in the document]
simao: @houcros it's equivalent but yes I'll change it

simao:
@jerrytrieu  this is ready to go on my side

simao:
@jerrytrieu  can we merge?

jerrytrieu:
Still need to wait for UI to fully switch to new campaign/update.

simao:
@houcros please check last commit

src/test/scala/com/advancedtelematic/campaigner/db/CampaignsSpec.scala
```diff
@@ -105,7 +105,7 @@ class CampaignsSpec extends AsyncFlatSpec
   }
 
   "finishing devices" should "work with one campaign" in {
-    val group    = GroupId.generate()
+    val _    = GroupId.generate()
```
simao: then you don't need this line?

houcros: Right 😆 I removed it now.

PR #63, opened 2018-08-24
Return update objects as part of a response if there is conflict.

koshelev:
It is needed in case of external updates.

Use Case: external system tries to create an update, gets conflict and updates name / description if they are changed.


src/main/scala/com/advancedtelematic/campaigner/http/UpdateResource.scala
```diff
@@ -54,6 +53,26 @@ class UpdateResource(extractNamespace: Directive1[Namespace], deviceRegistry: De
 
   val groupUpdateResolver = new GroupUpdateResolver(deviceRegistry, resolver)
 
+  import Errors.ConflictingUpdate
+
+  private[this] def duplicateUpdatesHanlder(namespace: Namespace, externalUpdateId: ExternalUpdateId) = {
```
simao: s/Hanlder/Handler/

src/main/scala/com/advancedtelematic/campaigner/http/UpdateResource.scala
```diff
@@ -54,6 +53,26 @@ class UpdateResource(extractNamespace: Directive1[Namespace], deviceRegistry: De
...
+    import io.circe.syntax._
+    ExceptionHandler {
+      case `ConflictingUpdate` =>
+        onSuccess(updateRepo.findByExternalIds(namespace, Seq(externalUpdateId))) { xs =>
+          xs.headOption match {
```
simao: think this could be `case x :+ _ =>` instead of headOption

PR #64, opened 2018-08-29
Add campaign status endpoints

simao:


simao:
Based on #62 so I don't have to rebase, see last 2 commits

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -38,30 +41,16 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
-      _ <- cancelTaskRepo.cancel(campaign)
-      _ <- groupStatsRepo.cancel(campaign)
-    } yield ()
+  def completeBatch(campaignId: CampaignId, group: GroupId, stats: Stats): Future[Unit] = db.run {
+    campaignStatusTransition.completeBatch(campaignId, group, stats).transactionally
+  }
 
-  def failedDevices(ns: Namespace, campaign: CampaignId): Future[Set[DeviceId]] = db.run {
-    campaignRepo.findAction(ns, campaign).flatMap { _ =>
+  def completeGroup(campaign: CampaignId, group: GroupId, stats: Stats): Future[Unit] = db.run {
```
houcros: Sorry, nitpicking here, but maybe name `campaignId` for similarity with the previous method? Must be my OCD kicking in 😆

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -150,25 +156,88 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
+protected [db] class CampaignStatusTransition(implicit db: Database, ec: ExecutionContext) extends CampaignSupport
+  with GroupStatsSupport with CancelTaskSupport  {
+
+  def devicesFinished(campaignId: CampaignId): DBIO[Unit] =
+    updateToCalculatedStatus(campaignId)
+
+  def launch(campaignId: CampaignId): DBIO[CampaignId] =
+    campaignRepo.setStatusAction(campaignId, CampaignStatus.launched)
+
+  def cancel(campaignId: CampaignId): DBIO[Unit] =
```
houcros: Maybe `cancelAction` for consistency with all the other `cancelAction`s ?

simao: This all seems rather arbitrary but I don't think it is. *Action methods are only used internally in this class, because this class only exposes DBIO Actions, it never exposes `Future[_]`.  

Where I think the confusion comes from, there are *Action methods for repositories, but those classes expose both DBIO and Future, so the method names have *Action to differentiate from the methods that expose Future.

houcros: Oh I see. Yes, that was were my confusion was coming from. Thanks for the clarification :)

simao:
Closing in favor of #67

PR #65, opened 2018-09-03
Get update by id

koshelev:
Next step would be to also fetch it as part of Campaign, after #62 gets merged


src/main/scala/com/advancedtelematic/campaigner/http/UpdateResource.scala
```diff
@@ -74,6 +75,9 @@ class UpdateResource(extractNamespace: Directive1[Namespace], deviceRegistry: De
   val route: Route =
     extractNamespace { ns =>
       pathPrefix("updates") {
+        (path( PathMatchers.JavaUUID.map(UpdateId.apply) ) & pathEnd) { updateUuid =>
```
houcros: I think you can use `UpdateId.Path` from `com.advancedtelematic.libats.http.UUIDKeyAkka` here?

PR #66, opened 2018-09-04
Add support for custom log level for ats classes

simao:


src/main/resources/logback.xml
```diff
@@ -1,12 +1,13 @@
 <configuration>
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>%.-1level|%d{HH:mm:ss.SSS}|%logger{36}|%message%n%exception</pattern>
+            <pattern>%.-1level|%date{ISO8601, UTC}|%logger{36}|%message%n%exception</pattern>
```
simao: @alexhumphreys I changed this format to have the date of the line, as containers are running for multiple days, but I am worried that logback will no longer parse this properly, but IIRC it uses the date from nginx and not from this line?

PR #68, opened 2018-09-04
Simplify error handler

simao:


src/main/scala/com/advancedtelematic/campaigner/http/UpdateResource.scala
```diff
@@ -91,18 +76,23 @@ class UpdateResource(extractNamespace: Directive1[Namespace], deviceRegistry: De
...
-                  val resourceUri = req.uri.withPath(req.uri.path / uuid.uuid.toString)
-                  complete((StatusCodes.Created, List(Location(resourceUri)), uuid))
+          (post & entity(as[CreateUpdate]) & extractRequest & extractLog) { (createUpdate, req, log) =>
+            onComplete(updateRepo.persist(createUpdate.mkUpdate(ns))) {
+              case Success(uuid) =>
+                val resourceUri = req.uri.withPath(req.uri.path / uuid.uuid.toString)
+                complete((StatusCodes.Created, List(Location(resourceUri)), uuid))
+              case Failure(ConflictingUpdate) =>
+                onSuccess(updateRepo.findByExternalIds(ns, Seq(createUpdate.updateSource.id))) {
+                  case x +: _ =>
```
houcros: What is the meaning of this syntax while pattern-matching? Earlier today I couldn't find any docs about it and I'm curious :)

simao: This is the same as in with a list `x :: xs`, but because the value is a Seq and not  a List you need to use `+:` because `::` is not defined on Seq, only `+:`.

houcros: Ah I see, thanks :)

koshelev: Is definitely an improvement. Should we also extract the whole `onComplete` block into a function to have only routing logic inside of route? Otherwise routing will be lost in handling code if we add some more routed

PR #69, opened 2018-09-05
Use external resolver configured in user profile.

houcros:


src/main/scala/com/advancedtelematic/campaigner/Boot.scala
```diff
@@ -49,14 +52,14 @@ object Boot extends BootApp
 
   log.info(s"Starting $version on http://$host:$port")
 
-  val director = new DirectorHttpClient(directorUri)
-
   val deviceRegistry = new DeviceRegistryHttpClient(deviceRegistryUri)
+  val director = new DirectorHttpClient(directorUri)
+  val resolver = new ResolverHttpClient(userProfileUri)
```
koshelev: Why not `val userProfile = new UserProfileHttpClient(...)`?

src/main/scala/com/advancedtelematic/campaigner/client/ResolverClient.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  private def externalResolverUri(ns: Namespace): Future[Uri] = {
+    val path = uri.path / "api" / "v1" / "namespace_settings" / ns.get
+    val request = HttpRequest(HttpMethods.GET, uri.withPath(path))
+    execHttp[Uri](ns, request)
+  }
+
+  override def availableUpdatesFor(ns: Namespace, devices: Seq[DeviceId]): Future[Seq[ExternalUpdateId]] = {
+    externalResolverUri(ns).flatMap { uri =>
+      val entity = HttpEntity(ContentTypes.`application/json`, devices.asJson.noSpaces)
```
koshelev: Why not `GET /uri?ids=id1,id2,id3`?

houcros: I didn't know what requests the resolver accepts, so I made this assumption for now.
If that's the request it accepts I'll change it.

koshelev: I assumed it is in our design document, but seems like it is not. My implementation accepts the get as I described

simao: those params on URLs will have a shorter size limit, which is fine, but it means we need to account for that when fetching/batching. Or we could just ignore that for now as there will be only a 2 or 3 devices.

koshelev: AFAIK, query params have size limit only in browsers and there is no such restriction on http level

simao: It's not only in browsers, the http protocol itself doesn't have a limitation but in practice some clients/libraries do. But anyway this doesn't matter for now.

houcros: Updated to `GET` instead of `POST`.

src/main/scala/com/advancedtelematic/campaigner/http/UpdateResource.scala
```diff
@@ -53,8 +51,6 @@ class UpdateResource(extractNamespace: Directive1[Namespace], deviceRegistry: De
     HypermediaResource(links, update)
   }
 
-  val groupUpdateResolver = new GroupUpdateResolver(deviceRegistry, resolver)
```
koshelev: Why not  using the variable?

houcros: Right, that got lost somewhen while refactoring. I brought it back now :)

src/main/scala/com/advancedtelematic/campaigner/http/UpdateResource.scala
```diff
@@ -41,9 +40,11 @@ object HypermediaResource {
 
 final case class HypermediaResource[T](links: Seq[HypermediaResource.Link], value: T)
 
-class UpdateResource(extractNamespace: Directive1[Namespace], deviceRegistry: DeviceRegistryClient, resolver: Resolver)
+class UpdateResource(extractNamespace: Directive1[Namespace], deviceRegistry: DeviceRegistryClient, resolver: ResolverClient)
                     (implicit db: Database, ec: ExecutionContext) extends Settings with UpdateSupport {
 
+  private val groupUpdateResolver = new GroupUpdateResolver(deviceRegistry, resolver)
```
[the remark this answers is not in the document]
koshelev: namespace is passed as a parameter to `availableUpdatesFor(ns: Namespace, devices: Seq[DeviceId])`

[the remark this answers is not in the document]
simao: and?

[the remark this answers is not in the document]
koshelev: Why do you need a different instance of resolver for every namespace? It is not parametrised on ns, or am I missing something?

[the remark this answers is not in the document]
simao: You need to query user profile to find out the url for the resolver.

src/main/scala/com/advancedtelematic/campaigner/Boot.scala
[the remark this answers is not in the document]
houcros: I started out with that approach, but then it came to the point where I would need to instantiate resolvers in `UpdateResource`, where we don't have `ActorSystem` or `Materializer`.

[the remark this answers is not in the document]
simao: you have both of those, but this is not really a matter of approach, because you are instantiating a resolver with a user profile uri, this will not work.

src/main/scala/com/advancedtelematic/campaigner/http/UpdateResource.scala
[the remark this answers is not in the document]
koshelev: Yes, but is noty it what 
```scala
override def availableUpdatesFor(ns: Namespace, devices: Seq[DeviceId]): Future[Seq[ExternalUpdateId]] = {
    externalResolverUri(ns).flatMap { uri =>
```
is doing?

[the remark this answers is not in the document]
simao: alright, my fault, I didn't even see that.

I don't think resolverclient should know anything about user service. ResolverClient should know about some external resolver.

We should have a user profile client that returns the url, which is then used when we call or instantiate the resolver client.

src/main/scala/com/advancedtelematic/campaigner/Boot.scala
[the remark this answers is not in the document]
simao: Ok, sorry, now I see it.

So, you need to create a separate user profile client in Boot, where you need to also instantiate a new Resolver client, which doesn't accept any uri. Then when you actually call resolver client you can pass along an uri.

houcros:
I updated to use two independent clients, one for user profile and one for resolver.

houcros:
Squashed the commits.

houcros:
Shall we merge this @koshelev?

PR #70, opened 2018-09-07
Use HttpClient from libats.

houcros:
Based on `feat/OTA-741/use-external-resolver-for-updates` so it needs to be merged after #69.
The only relevant commit is the last one.


src/main/scala/com/advancedtelematic/campaigner/client/DirectorClient.scala
```diff
@@ -71,7 +73,7 @@ class DirectorHttpClient(uri: Uri)
       method = HttpMethods.PUT,
       uri    = uri.withPath(path),
     )
```
koshelev: i would add `.withNs(ns)` here, not when `execute` is called

houcros: Indeed. Updated now.

houcros:
Rebased again on top of `feat/OTA-741/use-external-resolver-for-updates` after squashing the commits.

PR #72, opened 2018-09-10
Fix data before running migration

simao:
Bump libats, fixing boot when update id does not exist


simao:
Requires manual steps in staging db to run migrations again

PR #75, opened 2018-09-13
Sort updates and campaigns by name and creation time.

houcros:


simao:
Needs rebase, but also I think we should use the same ideas as for device registry.

src/main/scala/com/advancedtelematic/campaigner/data/Codecs.scala
```diff
@@ -61,4 +62,7 @@ object Codecs {
 
   implicit val uriEncoder: Encoder[Uri] = Encoder.encodeString.contramap(_.toString)
   implicit val uriDecoder: Decoder[Uri] = Decoder.decodeString.map(Uri.apply)
+
+  implicit val decoderSortBy: Decoder[SortBy] = Decoder.enumDecoder(SortBy)
```
simao: why do you need this?

houcros:
Refactored to use the same ideas as in device registry.

PR #77, opened 2018-09-20
Enforce at least one device group when creating a campaign.

houcros:


src/test/scala/com/advancedtelematic/campaigner/actor/CampaignSchedulerSpec.scala
```diff
@@ -27,13 +28,13 @@ class CampaignSchedulerSpec extends ActorSpec[CampaignScheduler] with Campaigner
...
-    val groups   = arbitrary[Set[GroupId]].sample.get
-    val campaign = createDbCampaignWithUpdate(groups = groups).futureValue
+    val groups   = arbitrary[NonEmptyList[GroupId]].sample.get
+    val campaign = createDbCampaignWithUpdate(maybeGroups = Some(groups)).futureValue
 
     val parent   = TestProbe()
 
     campaigns.scheduleGroups(campaign.id, groups).futureValue
-    groups.foreach { g => deviceRegistry.setGroup(g, arbitrary[Seq[DeviceId]].sample.get) }
+    groups.map{ g => deviceRegistry.setGroup(g, arbitrary[Seq[DeviceId]].sample.get) }
```
simao: why not keep foreach? foreach is usually used when the iteration is just for side effects

houcros: There is no `foreach` for `NonEmptyList` 😕

houcros:
@jerrytrieu do you think we're ready to merge this week? We haven't messed around with the campaigner for a while, should be stable by now?

jerrytrieu:
@houcros let's merge

PR #79, opened 2018-09-21
Return only internal updates if no resolver is configured.

houcros:


src/main/scala/com/advancedtelematic/campaigner/http/GroupUpdateResolver.scala
```diff
@@ -28,7 +28,7 @@ class GroupUpdateResolver(deviceRegistry: DeviceRegistryClient, resolver: Resolv
     }
 
     for {
-      devices <- fetchDevices(offset = 0, limit = 50)
```
koshelev: Can we please have names back? Improves readability IMO

houcros: Names are back :)

PR #80, opened 2018-09-24
Show updates for multiple groups

houcros:


houcros:
Based on #79.

koshelev:
@houcros could you check why TC is failing to build this PR?

src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
```diff
@@ -298,14 +298,19 @@ protected class UpdateRepository()(implicit db: Database, ec: ExecutionContext)
     findByExternalIdsAction(ns, Seq(id)).failIfNotSingle(Errors.MissingExternalUpdate(id))
   }
 
-  def all(ns: Namespace, sortBy: SortBy = SortBy.Name, nameContains: Option[String] = None): Future[Seq[Update]] = db.run {
```
koshelev: I would prefer to parametrize the `all` function on update type

src/main/scala/com/advancedtelematic/campaigner/http/GroupUpdateResolver.scala
```diff
@@ -14,21 +14,20 @@ import scala.concurrent.{ExecutionContext, Future}
...
-          FastFuture.successful(devicesInBatch)
-        else if (requests >= maxRequests)
-          FastFuture.failed(Errors.TooManyRequestsToRemote)
-        else
-          fetchDevices(offset + devicesInBatch.size.toLong, limit, requests + 1).map { devicesInBatch ++ _}
-      }
-    }
+    def fetchDevices(offset: Long, limit: Long, requests: Int = 1): Future[Set[DeviceId]] =
+      Future
+        .sequence(groups.map(deviceRegistry.devicesInGroup(ns, _, offset, limit)))
```
koshelev: I think we need something like
```scala
def fetchDevices(groupId: GroupId, offset: Long, limit: Long, requests: Int = 1): Future[Set[DeviceId]] = ???

for {
  devices <- Future.sequence(groups.flatMap(groupId => fetchDevices(groupId, 0, 50)))
 ...
```
The problem with your implementation is it tries to fetch devices from all groups until at least one if not fully fetched.

koshelev: `DeviceRegistry` returns a paginated result, so we can tell if there will be to many requests to a remote already after first response is received

src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
houcros: Sure, I've updated this on #79 now.

src/main/scala/com/advancedtelematic/campaigner/http/GroupUpdateResolver.scala
```diff
@@ -9,26 +9,26 @@ import com.advancedtelematic.libats.data.DataType.Namespace
...
-        else if (requests >= maxRequests)
-          FastFuture.failed(Errors.TooManyRequestsToRemote)
-        else
-          fetchDevices(offset + devicesInBatch.size.toLong, limit, requests + 1).map { devicesInBatch ++ _}
-      }
+    def fetchDevices(groupId: GroupId, offset: Long, limit: Long): Future[Seq[DeviceId]] = {
+      deviceRegistry
+        .devicesInGroup(ns, groupId, offset, limit)
+        .flatMap {
+          case Nil => FastFuture.successful(Seq.empty[DeviceId])
```
koshelev: Probably
```scala
  case ds if ds.size < limit => FastFuture.successful(ds)
  case ds => fetchDevices(groupId, offset + ds.size.toLong, limit).map(ds ++ _)
```

src/test/scala/com/advancedtelematic/campaigner/http/UpdateResourceSpec.scala
```diff
@@ -33,12 +33,11 @@ class UpdateResourceSpec extends CampaignerSpec with ResourceSpec with UpdateSup
...
     }
 
-  private def getUpdates(groupId: Option[GroupId] = None, nameContains: Option[String] = None): HttpRequest = {
-    val m = List("groupId" -> groupId, "nameContains" -> nameContains).collect{
-      case (k, Some(v: GroupId)) => k -> v.show
-      case (k, Some(v)) => k -> v.toString
-    }
-    Get(apiUri("updates").withQuery(Query(m.toMap))).withHeaders(header)
+  private def getUpdates(groups: Seq[GroupId] = Seq(), nameContains: Option[String] = None): HttpRequest = {
+    var m: Seq[(String, String)] = groups.map("groupId" -> _.show)
```
koshelev: scala```
    val m: Seq[(String, String)] = nameContains.map(x => Seq("nameContains" -> x)).getOrElse(Seq.empty) ++ groups.map("groupId" -> _.show) :+ ("limit", "200")
```
WDYT?

**Comment by houcros** [comment]
@koshelev I updated after our quick pair programming session :) Please have a look when you can.
Better to start with #79. In the end I kept both PRs to have a nice history, but once we merge #79 there will only be one commit to merge here.

File: src/main/scala/com/advancedtelematic/campaigner/http/UpdateResource.scala [hunk 5]
```diff
@@ -72,14 +74,22 @@ class UpdateResource(extractNamespace: Directive1[Namespace], deviceRegistry: De
     }
   }
 
-  private def getUpdates(ns: Namespace, offset: Long, limit: Long, sortBy: Option[SortBy], nameContains: Option[String]) =
+  private def availableUpdatesForGroups(ns: Namespace, groups: Set[GroupId]): Route = {
+    onComplete(getGroupUpdates(ns, groups)) {
+      case Success(r) => complete(r)
+      case Failure(_: TimeoutException) => complete((StatusCodes.EnhanceYourCalm, "It took too long to retrieve all the devices in the given groups."))
```
koshelev: I would prefer to return `504 Gateway Timeout` or `500 Internal Server Error` here with an appropriate error object as a payload. 4xx means processing failed because of clients error. It is definitely not a case here

houcros: Sure, updated.

PR #84, opened 2018-11-08
Save errors in database when campaign fails

simao:
Retry campaign when campaign fails

Do not retry if campaign failed too many times

This is an interim solution for OTA-848 while we cannot add more
states to not change the api for frontend.

We store the errors in a db table and use that to track how many
errors occurred and whether a campaign already failed too many times,
so it's not picked up by the scheduler again.

The actor system hierarchy had to be changed again... tbh I think this
should just be rewritten from scratch.


simao:
Deployed and tested in paris hat

houcros:
@simao I'm sorry, I deleted this branch by accident. I've restored it but I can't reopen the PR 😕 Sorry!

PR #87, opened 2018-12-04
Use SlickPagination from libats

jochenschneider:


src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
```diff
@@ -279,7 +280,7 @@ protected class CancelTaskRepository()(implicit db: Database, ec: ExecutionConte
   }
 }
 
-protected class UpdateRepository()(implicit db: Database, ec: ExecutionContext) {
+protected class UpdateRepository()(implicit db: Database, ec: ExecutionContext) extends SlickPagination {
```
simao: I think we'd be better avoiding inheritance and just importing SlickPagination and using the object methods

jochenschneider: Fixed. (Not sure that I would call extending a trait "inheritance".)

simao: Why not? The class gets a new type, how is it not inheritance? 

what if SlickPagination was `abstract class` would it then be inheritance?

jochenschneider: > Scala programmers "mix in" traits rather than inherit from them, because mixing in a trait has important differences from the multiple inheritance found in many other languages.
(https://www.artima.com/pins1ed/traits.html)

simao: yes it has differences from multiple inheritance, but it's still inheritance.

jochenschneider: Mixing in a trait to me is half way between inheritance and composition. Calling mixing in a trait "inheritance" blurs to me the distinction to inheriting from a class.

PR #91, opened 2019-01-18
Add endpoint to patch name and description for updates.

houcros:


src/main/scala/com/advancedtelematic/campaigner/http/UpdateResource.scala
```diff
@@ -97,8 +97,13 @@ class UpdateResource(extractNamespace: Directive1[Namespace], deviceRegistry: De
   val route: Route =
     extractNamespace { ns =>
       pathPrefix("updates") {
-        (path(UpdateId.Path) & pathEnd) { updateUuid =>
-          complete(updateRepo.findById(updateUuid))
+        path(UpdateId.Path) { updateUuid =>
+          get {
+            complete(updateRepo.findByIdUnsafe(updateUuid))
```
houcros: I'd like to make this `updateRepo.findById(ns, updateUuid)` but I don't know who's using it and if it might break something (?)..

houcros:
It seems we won't be needing this after all.

PR #92, opened 2019-02-08
Implement creating and listing child campaigns

nkly:
The PR is blocked on https://github.com/advancedtelematic/libats/pull/91 which is needed for returning a proper error when non-existing `parent_campaign_id` is given during campaign creation.

### DIFF
--- a/src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
@@ -30,9 +30,10 @@ object DataType {
     status: CampaignStatus,
     createdAt: Instant,
     updatedAt: Instant,
+    parentCampaignId: Option[CampaignId],
     autoAccept: Boolean = true
   )
-  
+
   final case class ExternalUpdateId(value: String) extends AnyVal
 
   final case class UpdateSource(id: ExternalUpdateId, sourceType: UpdateType)
@@ -61,6 +62,7 @@ object DataType {
   final case class CreateCampaign(name: String,
                                   update: UpdateId,
                                   groups: NonEmptyList[GroupId],
+                                  parentCampaignId: Option[CampaignId],
                                   metadata: Option[Seq[CreateCampaignMetadata]] = None,
                                   approvalNeeded: Option[Boolean] = Some(false))
   {
@@ -73,6 +75,7 @@ object DataType {
         CampaignStatus.prepared,
         Instant.now(),
         Instant.now(),
+        parentCampaignId,
         !approvalNeeded.getOrElse(false)
       )
     }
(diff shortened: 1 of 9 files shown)


houcros:
@nkly it seems it won't be so easy to get that PR into libats, so I pushed one commit so we can move on with this topic for now.

nkly:
@houcros Thank you. Sounds fine to me. Maybe you can also take a look at #93? We can merge #93 first, get a green build in this PR and merge it afterwards. WDYT?

houcros:
Yes, that makes sense to me. #93 looks fine to me overall. But I think my approval alone wouldn't be enough to merge it, would need one of the leads (?).

nkly:
Okay, I see. Thanks for taking a look at it anyway!

src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
```diff
@@ -97,7 +99,7 @@ protected [db] class DeviceUpdateRepository()(implicit db: Database, ec: Executi
       .map(_.status)
       .update(status)
       .flatMap {
-        case 0 => DBIO.failed(Errors.DeviceNotScheduled)
+        case 0 => DBIO.failed(DeviceNotScheduled)
```
jerrytrieu: Let's avoid prettying/formatting/whitespace changes in `feat`/`bugfix` PRs. Better to have separate `refactor` branches.

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
```diff
@@ -90,13 +93,15 @@ object DataType {
...
     updatedAt: Instant,
+    parentCampaignId: Option[CampaignId],
+    childCampaignIds: Set[CampaignId],
     groups: Set[GroupId],
     metadata: Seq[CreateCampaignMetadata],
     autoAccept: Boolean
   )
 
   object GetCampaign {
-    def apply(c: Campaign, groups: Set[GroupId], metadata: Seq[CampaignMetadata]): GetCampaign =
```
houcros: I was thinking, do we actually want to send the child campaigns? I don't think we'll need them. Only the child campaign stats when requesting the stats of the parent.

houcros:
Please let's have a look at #95 before we proceed to merge this.

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
nkly: Unfortunately I don't have enough knowledge to address this, but I think we might need them to actually get the stats for them. WDYT @jerrytrieu?

nkly:
Hey @houcros looks like you've overwritten the changes that prevented child campaigns from appearing in the main list of campaigns: https://github.com/advancedtelematic/campaigner/commit/b22fc782eafc4a8c77884bb666d2ce8ff2f0789e. Is it not required anymore?

houcros:
Sorry, seems it got lost yesterday while rebasing onto latest master 😕. We should add that back.

nkly:
No worries, that shouldn't be hard to fix. I trust you on this :) 
I believe in the future it would be nice to use additional pull requests for such changes so that they are visible and we don't step on each other's toes. WDYT?

houcros:
Opened #97.

PR #93, opened 2019-02-08
Refactor test data generators to improve tests stability 

nkly:


src/test/scala/com/advancedtelematic/campaigner/util/CampaignerSpec.scala
```diff
@@ -2,14 +2,15 @@ package com.advancedtelematic.campaigner.util
...
 import com.advancedtelematic.libats.test.LongTest
 import org.scalacheck.Gen
+import org.scalacheck.rng.Seed
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.{FlatSpecLike, Matchers}
 
 trait CampaignerSpecUtil {
   implicit class GenerateOps[T](value: Gen[T]) {
-    def generate: T = value.sample.get
+    def generate: T = value.pureApply(Gen.Parameters.default, Seed.random(), retries = 100)
```
houcros: In device-registry we also have such a class:
```
object GeneratorOps {
  implicit class GenSample[T](gen: Gen[T]) {
    @tailrec
    final def generate: T =
      gen.sample match {
        case Some(v) => v
        case None    => generate
      }
  }
}
```
I don't know if there is any big advantage of one over the other, but the one from dev-reg is somewhat easier to read?
I guess at some point we can move this to libats.

src/test/scala/com/advancedtelematic/campaigner/util/CampaignerSpec.scala
```diff
@@ -2,14 +2,15 @@ package com.advancedtelematic.campaigner.util
...
 
-    def gen= generate
+    def gen = generate
```
houcros: Let's get rid of this?

src/test/scala/com/advancedtelematic/campaigner/util/CampaignerSpec.scala
nkly: Yeah, while it might be a bit easier to read, it has one disadvantage: it won't stop until it generates something. It shouldn't be a problem most of the time, but there is still a chance that it won't stop ever. `pureApply` on the other hand has a fixed limit of retries.

nkly:
Hey folks @houcros @jerrytrieu @simao. Since this PR is not changing the service functionality at all, rather making our developer's lives better, I believe it is okay to set a review deadline for it and merge it afterwards regardless of whether everybody had a chance to look at it. I'll do it on **Monday, Feb. 18** unless you have any objections. Please feel free to both comment on the code and share your thoughts about the review deadline practice. Thanks!

nkly:
Okay, stupid me realized that I don't have write access to the master branch. In that case you can simply ignore my previous message.

PR #94, opened 2019-02-12
Increase coverage

nkly:
Add several additional test cases. They increase statement/branch coverage from 85.6%/75% to 86.9%/83.3%

### DIFF
--- a/src/test/scala/com/advancedtelematic/campaigner/daemon/DeviceEventListenerSpec.scala
+++ b/src/test/scala/com/advancedtelematic/campaigner/daemon/DeviceEventListenerSpec.scala
@@ -8,10 +8,12 @@ import com.advancedtelematic.campaigner.data.DataType.{Campaign, DeviceStatus, D
 import com.advancedtelematic.campaigner.data.Generators._
 import com.advancedtelematic.campaigner.db.{Campaigns, DeviceUpdateSupport, UpdateSupport}
 import com.advancedtelematic.campaigner.util.{CampaignerSpec, DatabaseUpdateSpecUtil, FakeDirectorClient}
-import com.advancedtelematic.libats.messaging_datatype.DataType.{DeviceId, Event}
+import com.advancedtelematic.libats.data.DataType.Namespace
+import com.advancedtelematic.libats.messaging_datatype.DataType.{DeviceId, Event, EventType}
 import com.advancedtelematic.libats.messaging_datatype.Messages
 import com.advancedtelematic.libats.messaging_datatype.Messages.DeviceEventMessage
 import com.advancedtelematic.libats.test.DatabaseSpec
+import io.circe.Json
 import io.circe.syntax._
 import org.scalacheck.Arbitrary._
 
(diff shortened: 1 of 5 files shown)


src/test/scala/com/advancedtelematic/campaigner/daemon/DeviceEventListenerSpec.scala
```diff
@@ -30,6 +32,12 @@ class DeviceEventListenerSpec extends CampaignerSpec with DatabaseSpec with Devi
     Messages.DeviceEventMessage(campaign.namespace, event)
   }
 
+  private def genEventOfType(eventType: EventType): DeviceEventMessage = {
```
simao: if we are using Generators and not returning a `Gen[_]` then the method name should not start with `gen`

src/test/scala/com/advancedtelematic/campaigner/daemon/DeviceEventListenerSpec.scala
```diff
@@ -65,4 +73,14 @@ class DeviceEventListenerSpec extends CampaignerSpec with DatabaseSpec with Devi
 
     deviceUpdateRepo.findByCampaign(campaign.id, DeviceStatus.failed).futureValue shouldBe Set(device)
   }
+
+  it should "raise IllegalArgumentException if the message is of non-acceptable type" in {
+    val msg = genEventOfType(EventType("campaign_accepted", 42))
+    listener.apply(msg).failed.futureValue.isInstanceOf[IllegalArgumentException]
```
simao: `listener.apply(msg).failed.futureValue shouldBe a[IllegalArgumentException]`

src/test/scala/com/advancedtelematic/campaigner/db/CampaignStatusRecalculateSpec.scala
```diff
@@ -0,0 +1,121 @@
...
+    val groups = GroupWithDevices.genNonEmptyListOf.generate
+
+    for {
+      campaign <- createCampaignWithoutStatus(maybeGroups = Some(groups.map(_.id)))
+      _ <- insertGroupStatsFor(campaign.id, groups.toList)
+      _ <- groupStatsRepo.updateGroupStatsAction(
+        campaign.id, groups.head.id, GroupStatus.cancelled, Stats(0, 0)).run
+      _ <- new CampaignStatusRecalculate().run
+      updatedCampaign <- campaignRepo.find(campaign.id)
+    } yield {
```
simao: curly braces not needed for a single statement

simao: Same in all examples

src/test/scala/com/advancedtelematic/campaigner/db/CampaignStatusRecalculateSpec.scala
```diff
@@ -0,0 +1,121 @@
...
+import org.scalacheck.Gen
+import org.scalatest._
+import org.scalatest.concurrent.ScalaFutures
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import slick.jdbc.MySQLProfile.api._
+
+final class CampaignStatusRecalculateSpec
+    extends TestKit(ActorSystem("CampaignStatusRecalculateSpec"))
+    with AsyncFlatSpecLike
```
simao: If you really like this style, keep it, but this leads to rather weird testing, for comprehensions everywhere, beating non future values into futures, and always having to end a future with an assertion...

src/test/scala/com/advancedtelematic/campaigner/daemon/DeviceEventListenerSpec.scala
nkly: Good one :+1:

src/test/scala/com/advancedtelematic/campaigner/db/CampaignStatusRecalculateSpec.scala
nkly: Sure. Do you follow a particular code style in projects? Where can I take a look at it? Would be nice to know the rules before playing the game :)

src/test/scala/com/advancedtelematic/campaigner/db/CampaignStatusRecalculateSpec.scala
nkly: TBH I just spotted `AsyncFlatSpec` usage in one of the tests and just followed it without much research. Later I've discovered `ScalaFutures` mixin which looks much better IMO. I'll refactor this test.

src/test/scala/com/advancedtelematic/campaigner/daemon/DeviceEventListenerSpec.scala
nkly: Sure. Same question about code style as in a comment about `yield` braces applies.

src/test/scala/com/advancedtelematic/campaigner/db/CampaignStatusRecalculateSpec.scala
simao: We did at some point use scalastyle.xml but it caused more discussion than the time to discuss the changes ;) So we don't follow any particular code style except for the projects that do include scalastyle.xml

nkly: Okay, got it. Thanks!

PR #95, opened 2019-02-15
Implement creation of retry-campaigns.

houcros:
Maybe I'm beating a dead horse here, but yesterday @nkly and I realised about the implementation outline in OTA-2113 and I believe it would be better approach.
In particular the use of a new endpoint `POST /api/v2/campaigns/:campaignId/campaigns` would make life easier for the FE: they create a new group and only need that groupId to create the retry-campaign. The other way forces the FE to send the updateId and the name on top of the groupId. I think it also makes it easier to maintain integrity.
I just leave this here and let's discuss on Monday which way we should go.


houcros:
Alright so let's keep things simple here and move on with #92.

PR #96, opened 2019-02-15
[OTA-2194] Change the /campaigns endpoint to return campaigns instead of IDs

houcros:
Change the `/campaigns` endpoint so that:

- It return campaigns instead of campaign IDs.
- It sorts the campaigns from newest to oldest created by default.

*Note* Should be merged together with related FE changes, else campaigns will stop showing in the UI.


houcros:
Not needed anymore. Superseded by #104 and #106.

PR #98, opened 2019-02-22
(OTA-2255) Take cancelled devices into account when calculating campaign status

nkly:
Fixes OTA-2255

Previously, when a device is individually cancelled, it is not taken
into account when calculating the campaign status, thus leaving the
campaign in 'launched' state indefinitely (`N_finished` is always
less than `N_affected`, and `N_cancelled` does not take individually
cancelled devices into account, see `CampaignStatusTransition#calculateCampaignStatus` method). Now such devices are counted as 'finished', thus allowing the campaign to
progress towards 'finished' state. When the whole campaign is cancelled,
however, it will end up in 'cancelled' states as it was before.

Signed-off-by: Nikolai Obedin <nikolai.obedin@here.com>

### DIFF
--- a/src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
@@ -233,7 +233,7 @@ protected [db] class CampaignStatusTransition(implicit db: Database, ec: Executi
     val countFinishedDevices =
       Schema.deviceUpdates
         .filter(_.campaignId === campaign).map(_.status)
-        .filter(status => status === DeviceStatus.successful || status === DeviceStatus.failed)
+        .filter(status => status === DeviceStatus.successful || status === DeviceStatus.failed || status === DeviceStatus.cancelled)
         .length.result
 
     for {
--- a/src/test/scala/com/advancedtelematic/campaigner/http/CampaignResourceSpec.scala
+++ b/src/test/scala/com/advancedtelematic/campaigner/http/CampaignResourceSpec.scala
@@ -23,13 +23,15 @@ import org.scalacheck.Arbitrary._
 import org.scalacheck.Gen
 import org.scalactic.source
 import org.scalatest._
+import org.scalatest.prop.PropertyChecks
 
 class CampaignResourceSpec
     extends CampaignerSpec
     with ResourceSpec
     with CampaignSupport
     with UpdateResourceSpecUtil
-    with GivenWhenThen {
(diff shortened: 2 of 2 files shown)


nkly:
Not sure that this is the correct fix. Maybe it makes sense to count such devices together with the cancelled groups. That would make, however, the whole campaign to get `cancelled` status, which might not be relevant if, for example, only one device from the large fleet was cancelled.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -233,7 +233,7 @@ protected [db] class CampaignStatusTransition(implicit db: Database, ec: Executi
     val countFinishedDevices =
```
nkly: As @houcros [mentioned](https://github.com/advancedtelematic/campaigner/pull/101/files/5c6f04bfeccb7bc5e0c65d216dde873041e1c590#r2598533290):

Something is odd. If we make this modification, then `val countFinishedDevices` is misleading, because "finished" is defined as "either successful or failed", as in the method `countFinished` up in this file. I feel either this modification is not needed or the name should be changed, but I'm not sure...

nkly: I think the modification makes sense, but the name becomes misleading, that's true. I will also rename other variables to reflect better whether a number is coming from a group stats or device updates table.

src/test/scala/com/advancedtelematic/campaigner/http/CampaignResourceSpec.scala
```diff
@@ -354,4 +355,40 @@ class CampaignResourceSpec
     }
   }
 
+  "GET /campaigns/:id/stats" should "return correct statistics" in {
+    val campaigns = Campaigns()
+    val groupId = genGroupId.generate
+
+    val successDevices = Gen.listOf(genDeviceId).generate
```
koshelev: Why are we generating a single sample?

nkly: I'm not sure I understood you. Is your question "why do we only test one campaign case and not make it into a property-based test instead"? If so, then there is no good reason for it indeed and I can gladly convert it into a property-based test.

koshelev: yes, that is what i mean. Let's do it!

PR #99, opened 2019-02-22
[OTA-2253] Create failure groups in dev-reg when receiving installation results.

houcros:
When implementing retry campaigns, we need some way to keep track of the group of failed devices that are associated with each retry campaign. This is the solution we thought of with @nkly.

### DIFF
--- a/src/main/resources/db/migration/V12__new_table_failed_groups.sql
+++ b/src/main/resources/db/migration/V12__new_table_failed_groups.sql
@@ -1,0 +1,8 @@
+CREATE TABLE failed_groups (
+  campaign_id  CHAR(36) NOT NULL,
+  group_id     CHAR(36) NOT NULL,
+  failure_code VARCHAR(200) NOT NULL,
+
+  CONSTRAINT pk_failed_groups PRIMARY KEY (campaign_id, failure_code),
+  CONSTRAINT fk_failed_groups_campaign_id FOREIGN KEY (campaign_id) REFERENCES campaigns (uuid)
+);
--- a/src/main/scala/com/advancedtelematic/campaigner/daemon/DeviceInstallationReportListener.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/daemon/DeviceInstallationReportListener.scala
@@ -1,17 +1,20 @@
 package com.advancedtelematic.campaigner.daemon
 
+import akka.http.scaladsl.util.FastFuture
 import cats.syntax.show._
+import com.advancedtelematic.campaigner.client.DeviceRegistryClient
 import com.advancedtelematic.campaigner.data.DataType._
 import com.advancedtelematic.campaigner.db.{Campaigns, UpdateSupport}
 import com.advancedtelematic.campaigner.http.Errors
-import com.advancedtelematic.libats.data.DataType.{CampaignId => CampaignCorrelationId, MultiTargetUpdateId}
+import com.advancedtelematic.libats.data.DataType.{MultiTargetUpdateId, Namespace, CampaignId => CampaignCorrelationId}
+import com.advancedtelematic.libats.messaging_datatype.DataType.DeviceId
 import com.advancedtelematic.libats.messaging_datatype.Messages.DeviceInstallationReport
 import org.slf4j.LoggerFactory
 import slick.jdbc.MySQLProfile.api._
 
 import scala.concurrent.{ExecutionContext, Future}
 
-class DeviceInstallationReportListener()(implicit db: Database, ec: ExecutionContext)
+class DeviceInstallationReportListener(deviceRegistryClient: DeviceRegistryClient)(implicit db: Database, ec: ExecutionContext)
   extends (DeviceInstallationReport => Future[Unit]) with UpdateSupport {
 
   private lazy val _log = LoggerFactory.getLogger(this.getClass)
@@ -19,14 +22,19 @@ class DeviceInstallationReportListener()(implicit db: Database, ec: ExecutionCon
   val campaigns = Campaigns()
 
   def apply(msg: DeviceInstallationReport): Future[Unit] = {
-    val resultStatus = if (msg.result.success)
-      DeviceStatus.successful
-    else
-      DeviceStatus.failed
+    val success = msg.result.success
+    val resultStatus = if (success) DeviceStatus.successful else DeviceStatus.failed
 
     val f = msg.correlationId match {
       case CampaignCorrelationId(uuid) =>
-        campaigns.finishDevices(CampaignId(uuid), Seq(msg.device), resultStatus)
+        val campaignId = CampaignId(uuid)
+        val addDeviceToFailedDeviceGroupFuture =
+          addDeviceToFailedDeviceGroup(msg.namespace, campaignId, msg.result.code, msg.device)
+        for {
+          _ <- campaigns.finishDevices(CampaignId(uuid), Seq(msg.device), resultStatus)
+          isMainCampaign <- campaigns.findClientCampaign(campaignId).map(_.parentCampaignId.isEmpty)
+          _ <- if (isMainCampaign && !success) addDeviceToFailedDeviceGroupFuture else Future.unit
+        } yield ()
       case MultiTargetUpdateId(uuid) => for {
         update <- updateRepo.findByExternalId(msg.namespace, ExternalUpdateId(uuid.toString))
         _ <- campaigns.finishDevice(update.uuid, msg.device, resultStatus)
@@ -40,4 +48,19 @@ class DeviceInstallationReportListener()(implicit db: Database, ec: ExecutionCon
         _log.info(s"Got DeviceUpdateReport for device ${msg.device.show} which is not scheduled by campaigner, ignoring this message.")
     }
   }
+
+  private def addDeviceToFailedDeviceGroup(ns: Namespace, campaignId: CampaignId, failureCode: String, deviceId: DeviceId): Future[Unit] =
+    findOrCreateFailedDeviceGroup(ns, campaignId, failureCode)
+      .flatMap(gid => deviceRegistryClient.addDeviceToGroup(ns, gid, deviceId))
+
+  private def findOrCreateFailedDeviceGroup(ns: Namespace, campaignId: CampaignId, failureCode: String): Future[GroupId] =
+    campaigns.fetchFailedGroupId(campaignId, failureCode).flatMap {
+      case Some(g) =>
+        FastFuture.successful(g)
+      case None =>
+        deviceRegistryClient
+          .createGroup(ns, s"failed-group-campaignId-$campaignId-failureCode-$failureCode")
+          .flatMap(gid => campaigns.persistFailedGroup(campaignId, gid, failureCode))
+    }
+
 }
--- a/src/main/scala/com/advancedtelematic/campaigner/DaemonBoot.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/DaemonBoot.scala
@@ -43,7 +43,7 @@ object DaemonBoot extends BootApp
     "campaign-supervisor"
   )
 
-  startListener[DeviceInstallationReport](new DeviceInstallationReportListener())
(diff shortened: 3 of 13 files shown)


nkly:
For the reference: https://saeljira.it.here.com/browse/OTA-2253

src/main/resources/db/migration/V12__new_table_failed_groups.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE failed_groups (
+  campaign_id  CHAR(36) NOT NULL,
+  group_id     CHAR(36) NOT NULL,
+  failure_code VARCHAR(200) NOT NULL,
```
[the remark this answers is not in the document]
nkly: Actually, it should be the PK because we'd like to have only one device group per camapign & failure code.

src/main/resources/db/migration/V12__new_table_failed_groups.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE failed_groups (
```
nkly: Do you think that `failed_device_groups` would be a more precise name? Applies to class/method names below as well.

src/main/scala/com/advancedtelematic/campaigner/daemon/DeviceInstallationReportListener.scala
```diff
@@ -40,4 +45,24 @@ class DeviceInstallationReportListener()(implicit db: Database, ec: ExecutionCon
         _log.info(s"Got DeviceUpdateReport for device ${msg.device.show} which is not scheduled by campaigner, ignoring this message.")
     }
   }
+
+  private def addDeviceToFailedGroup(msg: DeviceInstallationReport): Unit =
+    if (!msg.result.success) {
+      val failureCode = msg.result.code
+      msg.correlationId match {
+        case MultiTargetUpdateId(_) =>
```
nkly: Shall we also handle this case in a manner it is done [here](https://github.com/advancedtelematic/campaigner/blob/8bdac1530aa081d5c0bd1963e644a715c453f420/src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala#L43)?

src/main/resources/db/migration/V12__new_table_failed_groups.sql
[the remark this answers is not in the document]
houcros: Indeed, but as I see it the whole triplet is unique right? For a group and campaign there is one failure code, and for a group and failure code there is one campaign (unless we intend to try to reuse groups, which seems unreasonable). Hence the PK definition here.

src/main/resources/db/migration/V12__new_table_failed_groups.sql
houcros: On the one hand yes, I think it would be more precise. On the other hand, all over dev-reg and campaigner there is just the "group" concept to represent "device group". So it would kinda break the convention I guess. But I have no strong opinion about this, I can rename it.

src/main/scala/com/advancedtelematic/campaigner/daemon/DeviceInstallationReportListener.scala
houcros: I think we can safely ignore this branch, because we really just care about what happens with campaigns. AFAIK the `MultiTargetUpdateId` is used when you push an update for a single device from the DDV (device detailed view). And that doesn't mingle with our US.

src/main/scala/com/advancedtelematic/campaigner/daemon/DeviceInstallationReportListener.scala
```diff
@@ -40,4 +45,24 @@ class DeviceInstallationReportListener()(implicit db: Database, ec: ExecutionCon
         _log.info(s"Got DeviceUpdateReport for device ${msg.device.show} which is not scheduled by campaigner, ignoring this message.")
     }
   }
+
+  private def addDeviceToFailedGroup(msg: DeviceInstallationReport): Unit =
```
nkly: The code works perfectly fine, but the name of the function might be a bit misleading. Let's try to come up with a better one by writing down what does the function do. It takes a device installation report and:

- checks that the installation failed
- extracts the failure code
- extracts the campaign id
- fetches the failed device group _or_ creates one if nothing is found
- finally, adds a device to the group (actually doing what is stated in the name)

Whew, that's a list! So, from here we can go two ways: either create the most specific name that won't mislead anybody (for example, `parseInstallationReportAndAddDeviceToANewOrExistingFailedGroupIfItIsFailed`) or use a generic name so that it won't disclose any specifics (for example, `processInstallationReport`). I personally find neither of those options to be fine.

What else we can do? Let's take a look at the code around. One thing that strucks me is that we already parse the report in `apply` method, processing it differently based on the correlation ID. That might be a good place for us to step in and add a processing step for a failed device. Another thing would be to split the `addDeviceToFailedGroup` function into a smaller methods that compose well. For example, one method could perform the FDG lookup and creation, another one will add the device to the group returned by the previous method. The latter could be used in `apply` directly. Yet another idea is to pass to these utility method only the minimum of necessary information and not drag the whole report around. Following these steps, we can come up with a code that looks similar to this one:

```scala
  def apply(msg: DeviceInstallationReport): Future[Unit] = {
    val resultStatus = if (msg.result.success)
      DeviceStatus.successful
    else
      DeviceStatus.failed

    val f = msg.correlationId match {
      case CampaignCorrelationId(uuid) =>
        val campaignId = CampaignId(uuid)
        val addDeviceToFailedDeviceGroupFuture =
          addDeviceToFailedDeviceGroup(msg.namespace, campaignId, msg.result.code, msg.device)
        for {
          _ <- if (!msg.result.success) addDeviceToFailedDeviceGroupFuture else Future.unit
          _ <- campaigns.finishDevices(CampaignId(uuid), Seq(msg.device), resultStatus)
        } yield ()
      case MultiTargetUpdateId(uuid) => for {
        update <- updateRepo.findByExternalId(msg.namespace, ExternalUpdateId(uuid.toString))
        _ <- campaigns.finishDevice(update.uuid, msg.device, resultStatus)
      } yield ()
    }

    f.recover {
      case Errors.MissingExternalUpdate(_) =>
        _log.info(s"Could not find an update with external id ${msg.correlationId}, ignoring message")
      case Errors.DeviceNotScheduled =>
        _log.info(s"Got DeviceUpdateReport for device ${msg.device.show} which is not scheduled by campaigner, ignoring this message.")
    }
  }

  private def addDeviceToFailedDeviceGroup(namespace: Namespace, campaignId: CampaignId, failureCode: String, deviceId: DeviceId): Future[Unit] =
    findOrCreateFailedDeviceGroup(namespace, campaignId, failureCode)
      .flatMap(gid => deviceRegistryClient.addDeviceToGroup(namespace, gid, deviceId))

  private def findOrCreateFailedDeviceGroup(namespace: Namespace, campaignId: CampaignId, failureCode: String): Future[GroupId] =
    campaigns.fetchFailedGroupId(campaignId, failureCode).flatMap {
      case Some(g) =>
        FastFuture.successful(g)
      case None =>
        deviceRegistryClient
          .createGroup(namespace, s"failed-group-campaignId-$campaignId-failureCode-$failureCode")
          .flatMap(gid => campaigns.persistFailedGroup(campaignId, gid, failureCode))
    }
```

Now we are (hopefully) more clear about our intentions and :drum: :drum: :drum: use less code!

nkly: Of course, if you believe that the code above could be improved even further, then go for it!

src/main/scala/com/advancedtelematic/campaigner/daemon/DeviceInstallationReportListener.scala
nkly: Okay, makes sense. Didn't know about the specifics of a multi-target update. Thank you!

src/main/resources/db/migration/V12__new_table_failed_groups.sql
nkly: Ah, true, it is only `group` indeed... Yeah, I guess in this case it's better to leave it as it is now.

src/main/resources/db/migration/V12__new_table_failed_groups.sql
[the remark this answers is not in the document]
nkly: Having the whole triplet to be unique allows creating several groups for the same campaign and failure code because triplets `(cid, gid1, fc)` and `(cid, gid2, fc)` are different, and thus, unique. On the other hand, when we have only campaign ID and failure code as a PK, the triplets above become the same, preventing us from erroneously inserting another group for the same campaign and failure code.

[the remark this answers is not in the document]
houcros: Damn, right! Fixed now.

src/main/scala/com/advancedtelematic/campaigner/daemon/DeviceInstallationReportListener.scala
houcros: Thanks, I like that suggestion :) I've added that code with some minor tweaks.

src/main/scala/com/advancedtelematic/campaigner/daemon/DeviceInstallationReportListener.scala
```diff
@@ -1,32 +1,39 @@
...
+    val success = msg.result.success
+    val resultStatus = if (success) DeviceStatus.successful else DeviceStatus.failed
 
     val f = msg.correlationId match {
       case CampaignCorrelationId(uuid) =>
-        campaigns.finishDevices(CampaignId(uuid), Seq(msg.device), resultStatus)
+        val campaignId = CampaignId(uuid)
+        val addDeviceToFailedDeviceGroupFuture =
+          addDeviceToFailedDeviceGroup(msg.namespace, campaignId, msg.result.code, msg.device)
+        for {
```
houcros: I just realised, this won't prevent us form creating groups for nested retry-campaigns... I think this should be:
```
for {
          isMainCampaign <- campaigns.findClientCampaign(campaignId).map(_.parentCampaignId.isEmpty)
          _ <- if (isMainCampaign && !success) addDeviceToFailedDeviceGroupFuture else Future.unit 
          _ <- campaigns.finishDevices(CampaignId(uuid), Seq(msg.device), resultStatus)
        } yield ()
```

nkly: Good catch :+1: I doubt, though, that a solution to it would be that simple. Ideally we should add/remove devices from FDGs whenever we receive an installation report. In the example you've given we should retrieve group ID from the retry campaign, remove device from that group and add to the group corresponding to the main campaign and a new failure code instead.

houcros: Right, it gets that tricky... But I think for our "simple" first implementation (one retry per failed group, one retry at the time, only devices that failed in the main campaign per failed group) this should be enough.

nkly: I'd be glad if it would, but since FE will depend on us returning correct FDGs, we need to keep the groups up-to-date :-/

houcros: But we don't return the FDGs, only how many devices there are in each FDG. And since for now we only allow to retry what failed in the original main campaign, we don't have to support devices changing of FDG after retrying, as long as we can decrease the counts when devices succeed after retry. And I think we can do that with that we have.

nkly: We can decrease the overall counters without updating FDG, that's true, but AFAIK FE will also show a percentage bar per each failed group. For that we need to keep the groups up-to-date, unless there is another way to calculate that percentage.

houcros:
Not needed with the new approach we've taken. See #104 and #106.

PR #100, opened 2019-02-22
[OTA-2270] New endpoint to retry devices failed with a failure group.

houcros:
Built on #99.

If we (the BE) are gonna create and store the failure groups (see #99), then I don't see a good reason to expose them to the FE. It might be better to abstract this. After all the FE wants to retry devices that failed with a given failure group, they're not interested in how that's implemented.
So I've added this endpoint where the FE only needs to send the failure code. We have the failure group and the main campaign in the BE and we can create the retry campaign.
Any thoughts on this?

### DIFF
--- a/src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
@@ -44,8 +44,7 @@ class CampaignResource(extractAuth: Directive1[AuthedNamespaceScope], director: 
           s"retryCampaignWith-mainCampaign-${mainCampaign.id.uuid}-failureCode-${request.failureCode}",
           mainCampaign.updateId,
           retryGroup,
-          Some(mainCampaign.id)
-        ).mkCampaign(ns)
+        ).mkRetryCampaign(ns, mainCampaign.id)
         campaigns.create(retryCampaign, retryGroup, Nil)
     }
   }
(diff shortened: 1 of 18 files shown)


nkly:
I think your reasons for the separate endpoint are valid and it's a good idea :+1:

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
```diff
@@ -79,14 +96,19 @@ class CampaignResource(extractAuth: Directive1[AuthedNamespaceScope], director:
...
-          (post & path("launch")) {
-            complete(campaigns.launch(campaign.id))
+          post {
+            path("launch") {
+              complete(campaigns.launch(campaign.id))
+            } ~
+            path("cancel") {
+              complete(campaigns.cancel(campaign.id))
+            } ~
+            (path("retry-failed") & entity(as[RetryFailedDevices])) { request =>
```
houcros: Maybe this is not the best name for the endpoint because we're not retrying yet, just creating a retry-campaign. Two ideas: rename the endpoint to reflect that, or create the retry-campaign and launch it already. In any case we still return the campaignId to the FE. Any preference?

houcros:
I reverted the changes to `POST /api/v2/campaings`. If we're gonna use this new endpoint to create retry-campaigns, we shouldn't use the former.

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
nkly: I'd vote for launching at the time of creation because it's gonna be done by FE anyway with a next request.

nkly:
Heads up: I'm gonna pick this PR and apply the changes on top of #104 and #102 to make sure the endpoint is there by the end of the sprint.

houcros:
Superseded by #106.

PR #101, opened 2019-02-25
(OTA-2114) Merge statistics of the main and all retry campaigns

nkly:
Depends on #98

### DIFF
--- a/src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
@@ -51,11 +51,13 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
     campaignStatusTransition.completeGroup(campaign, group, stats).transactionally
   }
 
-  def failedDevices(campaign: CampaignId): Future[Set[DeviceId]] = db.run {
+  /**
+   * Given a campaign ID, returns IDs of all devices that are in `failed` state
+   */
+  private def findFailedDevicesAction(campaign: CampaignId): DBIO[Set[DeviceId]] =
     campaignRepo.findAction(campaign).flatMap { _ =>
       deviceUpdateRepo.findByCampaignAction(campaign, DeviceStatus.failed)
     }
-  }
 
   def freshCancelled(): Future[Seq[(Namespace, CampaignId)]] =
     cancelTaskRepo.findPending()
@@ -95,7 +97,7 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
    * Returns the number of devices that took part in the given campaigns and
    * finished, either successfully or with a failure.
    */
-  def countFinished(campaignIds: Set[CampaignId]): Future[Long] =
+  private def countFinishedAction(campaignIds: Set[CampaignId]): DBIO[Long] =
     campaignRepo.countDevices(campaignIds) { status =>
       status === DeviceStatus.successful || status === DeviceStatus.failed
     }
@@ -104,7 +106,7 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
    * Returns the number of devices that took part in the given campaigns, but
    * were cancelled before the update could be applied.
    */
-  def countCancelled(campaignIds: Set[CampaignId]): Future[Long] =
+  private def countCancelledAction(campaignIds: Set[CampaignId]): DBIO[Long] =
     campaignRepo.countDevices(campaignIds) { status =>
       status === DeviceStatus.cancelled
     }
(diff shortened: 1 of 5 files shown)


src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -91,13 +91,21 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
+  def countFinished(campaignIds: Set[CampaignId]): Future[Long] =
+    campaignRepo.countDevices(campaignIds) { status =>
       status === DeviceStatus.successful || status === DeviceStatus.failed
     }
 
-  def countCancelled(campaignId: CampaignId): Future[Long] =
-    campaignRepo.countDevices(campaignId) { status =>
+  /**
+   * Returns the number of devices that took part in the given campaigns, but
+   * were cancelled without being affected
```
houcros: `without being affected` could be misleading? If it was part of the campaign it was, by definition, affected (unless I misunderstood what are the affected devices).

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -155,16 +179,26 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
-        groupStatsRepo.findByCampaignAction(campaign)
-      }
-    }.map {
-      _.groupBy(_.group).map {
-        case (group, stats +: _) => (group, Stats(stats.processed, stats.affected))
-      }
-    }
+  /**
+   * Collects `processed` and `affected` counters for each group of the given
+   * campaign, sums the up, and returns campaign-wide numbers
```
houcros: Typo `sums them up`.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -155,16 +179,26 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
+   */
+  def campaignStatsFor(campaignId: CampaignId): Future[Stats] =
+    campaignStatsForAllOf(Set(campaignId))
+
+  /**
+   * Collects `processed` and `affected` counters for each group of each of the
+   * given campaigns, sums the up, and returns total numbers for all the
```
houcros: Ditto.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -233,7 +267,7 @@ protected [db] class CampaignStatusTransition(implicit db: Database, ec: Executi
     val countFinishedDevices =
```
houcros: Something is odd. If we make this modification, then `val countFinishedDevices` is misleading, because "finished" is defined as "either successful or failed", as in the method `countFinished` up in this file. I feel either this modification is not needed or the name should be changed, but I'm not sure...

src/test/scala/com/advancedtelematic/campaigner/http/CampaignResourceSpec.scala
```diff
@@ -354,4 +356,96 @@ class CampaignResourceSpec
...
+
+    When("requesting the stats for the main campaign")
+    Get(apiUri(s"campaigns/${mainCampaignId.show}/stats")).withHeaders(header) ~> routes ~> check {
+      status shouldBe OK
+
+      Then("the stats should be correct")
+      val campaignStats = responseAs[CampaignStats]
+      campaignStats.status shouldBe CampaignStatus.finished
+      campaignStats.finished shouldBe (mainCase.successfulDevices.size + mainCase.failedDevices.size)
+      campaignStats.failed should contain theSameElementsAs(mainCase.failedDevices)
```
houcros: Just some silly style tip: the parenthesis after `theSameElementAs` are not necessary if you want.

houcros:
For future reference, to follow the conventions, the branch should be named `(feat|bugfix|refactor)/<ticket-id>/<descriptive-name>`. This is specified in some wiki page that I can't find right now, so you'll have to  trust me on this one! 😄

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
nkly: This modification is coming from #98, so I'll move your comment to that PR.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -117,13 +125,29 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
+   * into account if any exist.
+   */
   def campaignStats(campaignId: CampaignId): Future[CampaignStats] = for {
-    status    <- findClientCampaign(campaignId).map(_.status)
-    finished  <- countFinished(campaignId)
-    cancelled <- countCancelled(campaignId)
-    failed    <- failedDevices(campaignId)
-    stats     <- campaignStatsFor(campaignId)
-  } yield CampaignStats(campaignId, status, finished, failed, cancelled, stats)
+    mainCampaign <- campaignRepo.find(campaignId)
```
koshelev: There is a possibility the result will be inconsistent, as every query is running in its own transaction. I would suggest to refactor the code to build a DB action running in a transaction.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -117,13 +125,29 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
+    retryCampaignIds <- campaignRepo.findChildIdsOf(campaignId)
+    Stats(mainProcessed, mainAffected) <- campaignStatsFor(campaignId)
+    Stats(retryProcessed, retryAffected) <- campaignStatsForAllOf(retryCampaignIds)
+    mainCancelled <- countCancelled(Set(campaignId))
```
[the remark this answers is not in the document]
nkly: The `Stats` object is coming from group stats, while `cancelled` is calculated using device stats. They could be put together, but that would require a lot of refactoring, which I tried to avoid for this PR. Looking at your later comment, though, it doesn't seem to be such a bad idea now. I'll fix the code to be executed in one transaction and see what I can do for this issue at the same time.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -117,13 +127,33 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
+  /**
+   * Calculates campaign-wide statistic counters, also taking retry campaings
+   * into account if any exist.
+   */
+  def campaignStats(campaignId: CampaignId): Future[CampaignStats] = db.run {
+    val statsAction = for {
+      mainCampaign <- campaignRepo.findAction(campaignId)
+      retryCampaignIds <- campaignRepo.findChildIdsOfAction(campaignId)
+      (mainProcessed, mainAffected) <- countProcessedAndAffectedDevicesForAllOfAction(Set(campaignId))
+      (retryProcessed, retryAffected) <- countProcessedAndAffectedDevicesForAllOfAction(retryCampaignIds)
```
koshelev: Why is group stats is not counting cancelled?

nkly: I don't know for certain, but it seems to me that counting cancelled devices using group stats is incorrect because of the ability to cancel individual devices. AFAIK it won't make it to the group stats.

PR #102, opened 2019-03-05
[OTA-2300] Store the result codes for each failed or successful device update.

houcros:
When parsing the installation reports, we now store the failure codes for the failed devices. We can later query to get the failed devices by failure code and use them to create a new retry-campaign.

https://saeljira.it.here.com/browse/OTA-2300

### DIFF
--- a/src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
@@ -92,23 +92,36 @@ protected [db] class DeviceUpdateRepository()(implicit db: Database, ec: Executi
       }
     }
 
-  protected [db] def setUpdateStatusAction(update: UpdateId, device: DeviceId, status: DeviceStatus): DBIO[Unit] =
+  /**
+    * Find all the device updates that failed with a given failure code.
+    */
+  def findFailedByFailureCode(campaignId: CampaignId, failureCode: String): Future[Set[DeviceId]] = db.run {
+    Schema.deviceUpdates
+      .filter(_.campaignId === campaignId)
+      .filter(_.status === DeviceStatus.failed)
+      .filter(_.resultCode === failureCode)
+      .map(_.deviceId)
+      .result
+      .map(_.toSet)
+  }
+
+  protected [db] def setUpdateStatusAction(update: UpdateId, device: DeviceId, status: DeviceStatus, resultCode: Option[String]): DBIO[Unit] =
     Schema.deviceUpdates
       .filter(_.updateId === update)
       .filter(_.deviceId === device)
-      .map(_.status)
-      .update(status)
+      .map(du => (du.status, du.resultCode))
+      .update((status, resultCode))
       .flatMap {
         case 0 => DBIO.failed(DeviceNotScheduled)
         case _ => DBIO.successful(())
       }.map(_ => ())
 
-  protected [db] def setUpdateStatusAction(campaign: CampaignId, devices: Seq[DeviceId], status: DeviceStatus): DBIO[Unit] =
+  protected [db] def setUpdateStatusAction(campaign: CampaignId, devices: Seq[DeviceId], status: DeviceStatus, resultCode: Option[String]): DBIO[Unit] =
     Schema.deviceUpdates
       .filter(_.campaignId === campaign)
       .filter(_.deviceId inSet devices)
-      .map(_.status)
-      .update(status)
+      .map(du => (du.status, du.resultCode))
+      .update((status, resultCode))
       .flatMap {
         case n if devices.length == n => DBIO.successful(())
         case _ => DBIO.failed(DeviceNotScheduled)
--- a/build.sbt
+++ b/build.sbt
@@ -26,7 +26,7 @@ libraryDependencies ++= {
   val akkaV = "2.5.13"
   val akkaHttpV = "10.1.1"
   val libatsV = "0.2.1-2-g962e326"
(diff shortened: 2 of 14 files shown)


src/main/resources/db/migration/V12__add_failure_code_failed_devices.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE device_updates ADD COLUMN failure_code VARCHAR(255) NULL;
```
nkly: Why is it safe to assume that failure code won't exceed 255 chars? Do we have a strict limitation on it somewhere?

src/main/scala/com/advancedtelematic/campaigner/daemon/DeviceInstallationReportListener.scala
```diff
@@ -19,17 +19,18 @@ class DeviceInstallationReportListener()(implicit db: Database, ec: ExecutionCon
...
   def apply(msg: DeviceInstallationReport): Future[Unit] = {
-    val resultStatus = if (msg.result.success)
-      DeviceStatus.successful
-    else
-      DeviceStatus.failed
+    val (resultStatus, failureCode) =
+      if (msg.result.success)
+        DeviceStatus.successful -> None
+      else
+        DeviceStatus.failed -> Some(msg.result.code)
```
nkly: What is in `msg.result.code` when an installation succeeds? Do you think it's worth storing this code in the table not only when an installation failed?

src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
```diff
@@ -92,23 +92,33 @@ protected [db] class DeviceUpdateRepository()(implicit db: Database, ec: Executi
...
+      .map(_.toSet)
+  }
+
+  protected [db] def setUpdateStatusAction(update: UpdateId, device: DeviceId, status: DeviceStatus, failureCode: Option[String]): DBIO[Unit] =
     Schema.deviceUpdates
       .filter(_.updateId === update)
       .filter(_.deviceId === device)
-      .map(_.status)
-      .update(status)
+      .map(du => du.status -> du.failureCode)
```
nkly: I'd personally prefer to use plain tuple `(du.status, du.failureCode)` instead of an arrow operator here as it looks more readable to me. According to http://slick.lightbend.com/doc/3.2.0/sql-to-slick.html#update, it seems to look more readable to Slick authors as well. It is a matter of taste, however, and since I'm not aware of any code style guidelines used in our projects, you are free to dismiss my comment.

src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
```diff
@@ -92,23 +92,33 @@ protected [db] class DeviceUpdateRepository()(implicit db: Database, ec: Executi
...
       }.map(_ => ())
 
-  protected [db] def setUpdateStatusAction(campaign: CampaignId, devices: Seq[DeviceId], status: DeviceStatus): DBIO[Unit] =
+  protected [db] def setUpdateStatusAction(campaign: CampaignId, devices: Seq[DeviceId], status: DeviceStatus, failureCode: Option[String]): DBIO[Unit] =
     Schema.deviceUpdates
       .filter(_.campaignId === campaign)
       .filter(_.deviceId inSet devices)
-      .map(_.status)
-      .update(status)
+      .map(du => du.status -> du.failureCode)
```
nkly: Ditto

src/main/scala/com/advancedtelematic/campaigner/daemon/DeviceInstallationReportListener.scala
koshelev: I would suggest to always store the code. I believe it was original idea that a code is always present and that is why the `success` flag was introduced - to distinguish between success and failure. I also can imagine different "success" codes

src/main/resources/db/migration/V12__add_failure_code_failed_devices.sql
houcros: If there is any limit specified I'm not aware of it. It just seemed to me 255 should be enough given the (few) uses I've seen. We could bump it up one or two orders of magnitude I guess? Or what would be a reasonable limit/type here? @jerrytrieu @koshelev

src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
houcros: Indeed, that seems more readable here. I will update it.

houcros: Done.

src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
houcros: Done.

src/main/scala/com/advancedtelematic/campaigner/daemon/DeviceInstallationReportListener.scala
houcros: As we discussed, I split the `finishDevice` method in three different methods to succeed/fail/cancel, and we save the `resultCode` when `DeviceStatus.success` or `DeviceStatus.failed`.

houcros:
Rebased.

houcros:
I rebased this on master, mistakingly I think. I should rebase it on #103. Will do after we merge it better.

houcros:
Ok, this is now rebased on later master. Nothing significant has changed but it's ready for another look.

PR #103, opened 2019-03-11
[OTA-1688] Listen for DeviceUpdateEvent

jerrytrieu:
DeviceInstallationReport is deprecated

Depends on https://github.com/advancedtelematic/director/pull/160

Signed-off-by: Jerry Trieu <jerry.trieu@here.com>

PR #104, opened 2019-03-14
[OTA-2305] Allow creating campaigns for a set of arbitrary devices

nkly:


nkly:
Hey folks! This is a WIP, but you can already peek at it and tell me your thoughs about the overall approach. I've decided to not create a separate table for linking devices to campaigns, but re-use `device_updates` instead by creating additional states such as `requested` (when device is initially added to the campaign) and `rejected` (when device is rejected by the Director and is not affected by the campaign). WDYT?

src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
```diff
@@ -3,76 +3,72 @@ package com.advancedtelematic.campaigner.actor
...
 
   def receive: Receive = {
-    case NextGroup =>
-      log.debug(s"next group")
-      campaigns.remainingGroups(campaign.id)
-        .map(_.headOption)
+    case NextBatch =>
+      log.debug("Requesting next batch")
+      campaigns.requestedDevices(campaign.id)
+        .map(result => BatchToSchedule(result.take(batchSize)))
```
houcros: Nitpicking: `.map(deviceIds => ...)` ?

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
```diff
@@ -168,9 +169,22 @@ object DataType {
     val prepared, launched, finished, cancelled = Value
   }
 
+  /**
```
houcros: AFAIU: all devices start as 'requested', which in the UI is shown as 'processed'. If it then moves to 'scheduled', the UI shows it as 'affected'. If it moves to 'rejected' it would be shown as 'not affected'.
I wonder if it could be worth to "document" that here. But on the other hand I'm reluctant, as that would couple the BE with the FE, and we might decide to display the stats in a different way in the future.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -169,8 +190,8 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
     io.transactionally
   }
 
-  def create(campaign: Campaign, groups: NonEmptyList[GroupId], metadata: Seq[CampaignMetadata]): Future[CampaignId] =
-    campaignRepo.persist(campaign, groups.toList.toSet, metadata)
+  def create(campaign: Campaign, groups: NonEmptyList[GroupId], devices: Set[DeviceId], metadata: Seq[CampaignMetadata]): Future[CampaignId] =
```
houcros: We could get rid of the groups here right? Since we're passing all the devices directly.

src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
```diff
@@ -175,13 +174,14 @@ protected [db] class GroupStatsRepository()(implicit db: Database, ec: Execution
...
 
-  def persist(campaign: Campaign, groups: Set[GroupId], metadata: Seq[CampaignMetadata]): Future[CampaignId] = db.run {
+  def persist(campaign: Campaign, groups: Set[GroupId], devices: Set[DeviceId], metadata: Seq[CampaignMetadata]): Future[CampaignId] = db.run {
     val f = for {
       _ <- (Schema.campaigns += campaign).recover {
         case Failure(UpdateFKViolation()) => DBIO.failed(MissingUpdateSource)
         case Failure(CampaignFKViolation()) => DBIO.failed(MissingParentCampaign)
       }.handleIntegrityErrors(ConflictingCampaign)
       _ <- Schema.campaignGroups ++= groups.map(campaign.id -> _)
+      < <- Schema.deviceUpdates ++= devices.map(did => DeviceUpdate(campaign.id, campaign.updateId, did, DeviceStatus.requested))
```
houcros: Typo `_` instead of `<`?

src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
```diff
@@ -175,13 +174,14 @@ protected [db] class GroupStatsRepository()(implicit db: Database, ec: Execution
...
   import com.advancedtelematic.libats.slick.db.SlickAnyVal._
 
-  def persist(campaign: Campaign, groups: Set[GroupId], metadata: Seq[CampaignMetadata]): Future[CampaignId] = db.run {
+  def persist(campaign: Campaign, groups: Set[GroupId], devices: Set[DeviceId], metadata: Seq[CampaignMetadata]): Future[CampaignId] = db.run {
     val f = for {
       _ <- (Schema.campaigns += campaign).recover {
         case Failure(UpdateFKViolation()) => DBIO.failed(MissingUpdateSource)
         case Failure(CampaignFKViolation()) => DBIO.failed(MissingParentCampaign)
       }.handleIntegrityErrors(ConflictingCampaign)
       _ <- Schema.campaignGroups ++= groups.map(campaign.id -> _)
```
houcros: Also here we could get rid of the groups and delete/archive the table?

src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
```diff
@@ -3,76 +3,72 @@ package com.advancedtelematic.campaigner.actor
...
+  } yield BatchComplete(affectedDevices, rejectedDevices)
 
   def receive: Receive = {
-    case NextGroup =>
-      log.debug(s"next group")
-      campaigns.remainingGroups(campaign.id)
-        .map(_.headOption)
+    case NextBatch =>
+      log.debug("Requesting next batch")
+      campaigns.requestedDevices(campaign.id)
```
houcros: Maybe `requestedDevices` could take a `limit` parameter and then we can limit at the DB level by passing the `batchSize`? I guess it won't make a difference right now, but at some point it could...

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
houcros: Yeah, actually we have `affected` and `processed` scattered through campaigner too, not just the FE. So do I think it wouldn't hurt to shortly describe that here. Otherwise I'm afraid this would get confusing.

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
```diff
@@ -55,6 +61,12 @@ class CampaignResource(extractAuth: Directive1[AuthedNamespaceScope], director:
       onSuccess(campaigns.findNamespaceCampaign(namespace, campaign)).flatMap(provide)
     }
 
+  private def fetchDevicesInGroups(ns: Namespace, groups: NonEmptyList[GroupId]): Future[Set[DeviceId]] = {
+    val groupSet = groups.toList.toSet
+    // TODO review the retrieval logic and 'limit' parameter
+    Future.sequence(groupSet.map(gid => deviceRegistry.devicesInGroup(ns, gid, 0, 50))).map(_.flatten)
```
houcros: I think this is a good use case for Future.traverse:
`Future.traverse(groupSet)(gid => deviceRegistry.devicesInGroup(ns, gid, 0, 50)).map(_.flatten)`.

src/main/resources/db/migration/V13__add_requested_status_to_device_updates_table.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE `device_updates`
```
houcros: We're planning to merge this after #103 right? Because in that case the migration should be V12.

nkly: Yep, that is the idea. I used V13 in order to avoid conflicts later.

nkly: Actually, after both #103 and #102 because they are ready and this PR is not yet.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
nkly: Not exactly. We return groups in `GetCampaign` and FE depends on this information. Luckily when they adapt to new stats, we can remove the field from `GetCampaign` and forget about groups completely.

houcros:
LGTM, nice clean-up!

src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
```diff
@@ -3,76 +3,73 @@ package com.advancedtelematic.campaigner.actor
...
-      batchSize,
-      campaign,
-      group),
-      s"group-scheduler-${group.show}"
-    )
+  private def schedule(deviceIds: Set[DeviceId]): Future[BatchComplete] = for {
+    affectedDevices <- deviceUpdateProcess.startUpdateFor(deviceIds.toSeq, campaign).map(_.toSet)
+    rejectedDevices = deviceIds -- affectedDevices
+    _ <- campaigns.rejectDevices(campaign.id, campaign.updateId, rejectedDevices.toSeq)
+    _ <- campaigns.updateStatus(campaign.id)
```
nkly: Move into separate function and run in transaction with `rejectDevices`, `acceptDevices`, `scheduleDevices`

src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
```diff
@@ -3,76 +3,73 @@ package com.advancedtelematic.campaigner.actor
...
+  } yield BatchComplete(affectedDevices, rejectedDevices)
 
   def receive: Receive = {
-    case NextGroup =>
-      log.debug(s"next group")
-      campaigns.remainingGroups(campaign.id)
-        .map(_.headOption)
+    case NextBatch =>
+      log.debug("Requesting next batch")
+      // TODO limit or stream device IDs from DB
```
nkly: Create separate issue or JIRA ticket for this and all the other todos

src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
```diff
@@ -227,14 +169,37 @@ protected class CampaignRepository()(implicit db: Database, ec: ExecutionContext
...
+    Schema.deviceUpdates.join(Schema.campaigns).on(_.campaignId === _.id)
+      .filter { case (deviceUpdate, _) => deviceUpdate.status === DeviceStatus.requested }
+      .map(_._2)
+      .result
+      .map(_.toSet)
+
+  /**
+   * Returns all campaigns that have all devices in `requested` state
+   */
+  def findAllNewlyCreated: DBIO[Set[Campaign]] = {
```
nkly: Give SQL a try, maybe it'll be more readable

src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
koshelev: Could you please add link to JIRA ticket to the comment?

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -222,45 +228,48 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
+    def persistDeviceUpdates(deviceIds: Set[DeviceId], status: DeviceStatus): DBIO[Unit] =
+      deviceUpdateRepo.persistManyAction(deviceIds.toSeq.map(deviceId =>
+        DeviceUpdate(campaign.id, campaign.updateId, deviceId, status)
+      ))
+
+    val action = for {
+      _ <- persistDeviceUpdates(acceptedDeviceIds, DeviceStatus.accepted)
+      _ <- persistDeviceUpdates(scheduledDeviceIds, DeviceStatus.scheduled)
+      _ <- persistDeviceUpdates(rejectedDeviceIds, DeviceStatus.rejected)
+      _ <- campaignStatusTransition.updateToCalculatedStatus(campaign.id)
```
koshelev: It would be interesting to check, what is a limit of a number of devices we can update in a single transaction

nkly: If we hit a perf bottlenect, we can always switch to SQL :)

src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
nkly: Ah, true, forgot about that. Will fix.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
koshelev: It will not be a perf bottleneck, the transaction will fail, I guess...

src/main/scala/com/advancedtelematic/campaigner/db/Repository.scala
```diff
@@ -127,74 +123,24 @@ protected [db] class DeviceUpdateRepository()(implicit db: Database, ec: Executi
         case _ => DBIO.failed(DeviceNotScheduled)
       }.map(_ => ())
 
-  def persistMany(updates: Seq[DeviceUpdate]): Future[Unit] = db.run {
-    DBIO.sequence(updates.map(Schema.deviceUpdates.insertOrUpdate)).transactionally.map(_ => ())
-  }
-}
-
-protected [db] class GroupStatsRepository()(implicit db: Database, ec: ExecutionContext) {
```
jerrytrieu: We lose group stats, but that feature is one on the roadmap, including showing failure codes by group. Okay for now.

PR #106, opened 2019-03-18
[OTA-2306] Create new endpoint to launch a retry-campaign

houcros:
Supersedes #100.

- Add a new endpoint to create and launch a new retry-campaign given a failure code.
- Store the failure code when creating the retry-campaign. We need that so that the FE can show the status of the retry-campaigns in the UI.

### DIFF
--- a/src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
@@ -30,6 +30,8 @@ object DataType {
     createdAt: Instant,
     updatedAt: Instant,
     mainCampaignId: Option[CampaignId],
+    // TODO ensure that they're both either Some or None
+    failureCode: Option[String],
     autoAccept: Boolean = true
   )
 
@@ -61,6 +63,7 @@ object DataType {
   final case class CreateCampaign(name: String,
                                   update: UpdateId,
                                   groups: NonEmptyList[GroupId],
+                                 // TODO remove mainCampaignId from here, we should use the other endpoint to launch a retry campaign.
                                   mainCampaignId: Option[CampaignId],
                                   metadata: Option[Seq[CreateCampaignMetadata]] = None,
                                   approvalNeeded: Option[Boolean] = Some(false))
@@ -75,6 +78,7 @@ object DataType {
         Instant.now(),
         Instant.now(),
         mainCampaignId,
+        None,
         !approvalNeeded.getOrElse(false)
       )
     }
@@ -87,9 +91,20 @@ object DataType {
                                        update: UpdateId,
                                        group: GroupId,
                                        mainCampaignId: CampaignId,
+                                       failureCode: String
                                       ) {
     def mkCampaign(ns: Namespace): Campaign =
-      CreateCampaign(name, update, NonEmptyList.one(group), Some(mainCampaignId)).mkCampaign(ns)
+      Campaign(
+        ns,
+        CampaignId.generate(),
+        name,
+        update,
+        CampaignStatus.prepared,
+        Instant.now(),
+        Instant.now(),
+        Some(mainCampaignId),
+        Some(failureCode)
+      )
   }
 
   final case class RetryFailedDevices(failureCode: String)
--- a/src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
@@ -30,7 +30,6 @@ object DataType {
     createdAt: Instant,
     updatedAt: Instant,
     mainCampaignId: Option[CampaignId],
-    // TODO ensure that they're both either Some or None
     failureCode: Option[String],
     autoAccept: Boolean = true
   )
@@ -63,8 +62,6 @@ object DataType {
   final case class CreateCampaign(name: String,
                                   update: UpdateId,
                                   groups: NonEmptyList[GroupId],
-                                 // TODO remove mainCampaignId from here, we should use the other endpoint to launch a retry campaign.
-                                  mainCampaignId: Option[CampaignId],
                                   metadata: Option[Seq[CreateCampaignMetadata]] = None,
                                   approvalNeeded: Option[Boolean] = Some(false))
   {
@@ -77,7 +74,7 @@ object DataType {
         CampaignStatus.prepared,
         Instant.now(),
         Instant.now(),
-        mainCampaignId,
+        None,
         None,
         !approvalNeeded.getOrElse(false)
       )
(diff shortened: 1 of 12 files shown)


src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
```diff
@@ -37,6 +40,31 @@ class CampaignResource(extractAuth: Directive1[AuthedNamespaceScope],
...
+    * such devices in `mainCampaign`, throw an error.
+    */
+  def retryFailedDevices(ns: Namespace, mainCampaign: Campaign, request: RetryFailedDevices): Future[Unit] =
+    campaigns.fetchFailedDevices(mainCampaign.id, request.failureCode).map(_.toSeq).flatMap {
+      case Nil =>
+        Future.failed(MissingFailedDevices(request.failureCode))
+
+      case deviceIds =>
+        // FIXME we just need to create this dummy group until we get totally rid of the groups.
+        val retryGroup = GroupId(UUID.fromString("0000-00-00-00-000000"))
```
nkly: The latest commits from #106 are loosen the groups type from `NonEmptyList` to just `Set`, so you won't need a fake group here and in `CreateRetryCampaign` class when you rebase on top of it.

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
```diff
@@ -83,6 +83,16 @@ object DataType {
       metadata.toList.flatten.map(_.toCampaignMetadata(campaignId))
   }
 
+  final case class CreateRetryCampaign(name: String,
+                                       update: UpdateId,
+                                       group: GroupId,
+                                       mainCampaignId: CampaignId,
+                                      ) {
+    def mkCampaign(ns: Namespace): Campaign =
+      CreateCampaign(name, update, NonEmptyList.one(group), Some(mainCampaignId)).mkCampaign(ns)
```
nkly: I'd simply copy over the code from `CreateCampaign::mkCampaign` method instead of re-using it because:
1. You won't need a `groupId` anymore
2. There is no complex logic in `CreateCampaign::mkCampaign` that we need to re-use

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
```diff
@@ -37,6 +40,31 @@ class CampaignResource(extractAuth: Directive1[AuthedNamespaceScope],
...
   }
 
+  /**
+    * Create and immediately launch a retry-campaign for all the devices that were processed by `mainCampaign` and
+    * failed with a code `failureCode`. We assume there is at least one failed device. Hence, if there were not any
+    * such devices in `mainCampaign`, throw an error.
+    */
+  def retryFailedDevices(ns: Namespace, mainCampaign: Campaign, request: RetryFailedDevices): Future[Unit] =
+    campaigns.fetchFailedDevices(mainCampaign.id, request.failureCode).map(_.toSeq).flatMap {
+      case Nil =>
```
nkly: Instead of converting the list of devices to and from `Seq`, I'd suggest to simply replace `case Nil` with `case devicesIds if devicesIds.isEmpty`. While it decreases readability a bit, it will:
1. Avoid unnecessary and potentially costly conversion
2. Remove non-explicit dependency on the standard Scala library implementation of `Set` being `List`. They might change it anytime to `Vector` and then your code will stop working because `Nil` won't ever match `Vector`.

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
nkly: Nevermind, just saw that you already did it.

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
nkly: Nevermind, just saw that you already did it

houcros:
Rebased again on #104.

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
houcros: Yeah I wasn't too happy about that myself. Thanks!

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
```diff
@@ -37,6 +40,25 @@ class CampaignResource(extractAuth: Directive1[AuthedNamespaceScope],
...
+    campaigns.fetchFailedDevices(mainCampaign.id, request.failureCode).flatMap {
+      case deviceIds if deviceIds.isEmpty =>
+        Future.failed(MissingFailedDevices(request.failureCode))
+
+      case deviceIds =>
+        // TODO OTA-2377 we just need to create this dummy group until we get totally rid of the groups.
+        val retryGroup = GroupId(UUID.fromString("0000-00-00-00-000000"))
+        val retryCampaign = CreateRetryCampaign(mainCampaign, request.failureCode).mkCampaign
+        campaigns
+          .create(retryCampaign, Set(retryGroup), deviceIds, Nil)
```
nkly: `Set.empty` could be used here. This way we don't need stub `retryGroup` anymore.

houcros: Indeed. Updated.

houcros:
@koshelev any feedback for this PR?

houcros:
@koshelev any thoughts here?

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
```diff
@@ -83,6 +84,22 @@ object DataType {
       metadata.toList.flatten.map(_.toCampaignMetadata(campaignId))
   }
 
+  final case class CreateRetryCampaign(mainCampaign: Campaign, failureCode: String) {
```
koshelev: Why do we need the case class?

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
```diff
@@ -37,6 +38,23 @@ class CampaignResource(extractAuth: Directive1[AuthedNamespaceScope],
     } yield campaignId
   }
 
+  /**
+    * Create and immediately launch a retry-campaign for all the devices that were processed by `mainCampaign` and
+    * failed with a code `failureCode`. We assume there is at least one failed device. Hence, if there were not any
+    * such devices in `mainCampaign`, throw an error.
+    */
+  def retryFailedDevices(ns: Namespace, mainCampaign: Campaign, request: RetryFailedDevices): Future[CampaignId] =
```
koshelev: I would suggest to replace this method and `CreateRetryCampaign` case-class with a `Campaigns.retryCampagn(ns, mainCampaign, errorCode)` method.

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
houcros: Removed.

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
houcros: Sure :) Updated as advised.

houcros:
@koshelev updated as suggested.

PR #108, opened 2019-03-21
Add `updated_at` field to `device_updates` table

nkly:
Might be in conflict with #106 because of the migration


houcros:
I say let's aim at merging this one before #106, since this one is lighter.

PR #109, opened 2019-03-22
[OTA-2384] Refactor `campaignStats` method

nkly:


src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -167,29 +137,40 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
       status = mainCampaign.status,
-      finished = mainFinished - (retryProcessed - retryAffected + retryCancelled),
+      finished = mainCnt.finished  - (retryCnt.rejected + retryCnt.cancelled),
       failed = failed,
-      cancelled = mainCancelled + retryCancelled,
-      processed = mainProcessed,
-      affected  = mainAffected - (retryProcessed - retryAffected)
+      cancelled = mainCnt.cancelled + retryCnt.cancelled,
+      processed = mainCnt.processed,
+      affected  = mainCnt.affected - retryCnt.rejected,
```
houcros: Why don't we send `succeeded` as well so we avoid all calculations in the FE?

nkly: This change is not a part of the PR. It is solely about refactoring the code.

houcros: Ah I see.

PR #110, opened 2019-03-22
[OTA-2390] Expose failures in main campaign stats

nkly:


src/test/scala/com/advancedtelematic/campaigner/db/CampaignsSpec.scala
```diff
@@ -63,3 +65,50 @@ class CampaignsSpec extends AsyncFlatSpec
...
+  with ScalaFutures
+  with CampaignSupport
+  with UpdateSupport
+  with DatabaseUpdateSpecUtil
+  with CampaignerSpecUtil {
+
+  import Arbitrary._
+
+  val campaigns = Campaigns()
+  implicit val defaultPatience = PatienceConfig(timeout = Span(2, Seconds))
```
houcros: If you want to get fancy: if you import `import org.scalatest.time.SpanSugar._` instead of `import org.scalatest.time.{Seconds, Span}` then you can write this as `PatienceConfig(timeout = 2 seconds)`.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -165,22 +175,52 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
         counts.getOrElse(DeviceStatus.failed, 0).toLong
     )
 
+    def processFailures(failedDevices: Set[DeviceUpdate], retryCampaigns: Set[Campaign]): Set[CampaignFailureStats] = {
+      val missingErrorCode = "MISSING_ERROR_CODE"
+      val retryStatusByFailureCode = retryCampaigns
+        .groupBy(_.failureCode.getOrElse(missingErrorCode))
+        .mapValues { campaigns =>
+          if (campaigns.exists(_.status == CampaignStatus.finished)) {
```
jerrytrieu: Only one campaign here?

houcros: For the current user story I think one campaign, yes. Because we can only launch one retry-campaign for failure code, hence grouping by failure code should be the same as grouping by campaign. If/when we generalise this, then we could have more than one campaign in each group.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -135,14 +128,31 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
     db.run(campaignRepo.findByUpdateAction(update))
 
+  /**
+   * Given a set of campaign IDs, finds all device updates that happened in
+   * these campaigns, selects the most recent failures and returns them.
+   */
+  protected[db] def findFailedDeviceUpdatesAction(campaignIds: Set[CampaignId]): DBIO[Set[DeviceUpdate]] = {
+    Schema.deviceUpdates
+      .join(Schema.deviceUpdates
+      .filter(upd => upd.campaignId.inSet(campaignIds))
```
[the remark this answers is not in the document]
nkly: How should it be indented?

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -135,14 +128,31 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
+        .groupBy(_.deviceId)
+        .map { case (id, upd) => (id, upd.map(_.updatedAt).max) })
+        .on { (fst, snd) => fst.deviceId === snd._1 && fst.updatedAt === snd._2 }
```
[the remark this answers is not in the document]
nkly: In plain SQL I'd use `rank()` over partition by `device_id` sorted by `update_at`, but I don't know how to express it in Slick DSL. Doing it in `sql`-escaped string would require duplicating (de)serialization in terms of `GetResult/SetParameter` implicit objects. Taking into account both of these things, I thought that we'll be fine with a `join` for now, and explore other options when the need arise. I'd appreciate your help if you know an alternative that's more efficient and elegant.

_Side note:_ I think that calculating stats upon a request is quite expensive when the numbers of campaigns and devices are sufficiently large. I'd do it at the time of receiving a device installation report and persist in a separate table, so that a result could be returned using simple `select`. Again, this measure would only be necessary when we have perf. issues with the current impl.

PR #111, opened 2019-03-28
[OTA-2308] Create endpoint to export campaign installation failures as CSV

houcros:
Based on #110. Only the last two commits are new here.

Arguably we want to move the export endpoint to campaigner. For that we're asking device-registry for the `oemId`s of the devices. But we don't store them, just put them in the CSV and forget about them.
If we move on with this, shortly after I'll remove the current export endpoint from device-registry.

### DIFF
--- a/src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
@@ -75,34 +75,34 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
   def markDevicesAccepted(campaign: CampaignId, update: UpdateId, devices: DeviceId*): Future[Unit] =
     deviceUpdateRepo.persistMany(devices.map { d => DeviceUpdate(campaign, update, d, DeviceStatus.accepted) })
 
-  def succeedDevice(updateId: UpdateId, deviceId: DeviceId, successCode: String): Future[Unit] =
-    finishDevice(updateId, deviceId, DeviceStatus.successful, Some(successCode))
+  def succeedDevice(updateId: UpdateId, deviceId: DeviceId, successCode: String, successDescription: String): Future[Unit] =
+    finishDevice(updateId, deviceId, DeviceStatus.successful, Some(successCode), Some(successDescription))
 
-  def succeedDevices(campaignId: CampaignId, devices: Seq[DeviceId], successCode: String): Future[Unit] =
-    finishDevices(campaignId, devices, DeviceStatus.successful, Some(successCode))
+  def succeedDevices(campaignId: CampaignId, devices: Seq[DeviceId], successCode: String, successDescription: String): Future[Unit] =
+    finishDevices(campaignId, devices, DeviceStatus.successful, Some(successCode), Some(successDescription))
 
-  def failDevice(updateId: UpdateId, deviceId: DeviceId, failureCode: String): Future[Unit] =
-    finishDevice(updateId, deviceId, DeviceStatus.failed, Some(failureCode))
+  def failDevice(updateId: UpdateId, deviceId: DeviceId, failureCode: String, failureDescription: String): Future[Unit] =
+    finishDevice(updateId, deviceId, DeviceStatus.failed, Some(failureCode), Some(failureDescription))
 
-  def failDevices(campaignId: CampaignId, devices: Seq[DeviceId], failureCode: String): Future[Unit] =
-    finishDevices(campaignId, devices, DeviceStatus.failed, Some(failureCode))
+  def failDevices(campaignId: CampaignId, devices: Seq[DeviceId], failureCode: String, failureDescription: String): Future[Unit] =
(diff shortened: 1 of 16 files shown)


src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -113,6 +113,19 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
   def fetchFailedDevices(campaignId: CampaignId, failureCode: String): Future[Set[DeviceId]] =
     deviceUpdateRepo.findFailedByFailureCode(campaignId, failureCode)
 
+  /**
+    * Returns all the devices that where targeted by the campaign with ID `campaignId` or any of its retry-campaigns
```
jerrytrieu: Typo `where`

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -113,6 +113,19 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
+    * and that are failed with code `failureCode` at the moment of the call, i.e. devices whose last installation
```
jerrytrieu: Can be better worded.

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
```diff
@@ -44,6 +47,32 @@ class CampaignResource(extractAuth: Directive1[AuthedNamespaceScope],
   def retryFailedDevices(ns: Namespace, mainCampaign: Campaign, request: RetryFailedDevices): Future[CampaignId] =
     campaigns.retryCampaign(ns, mainCampaign, request.failureCode)
 
+
+  implicit val installationFailureCsvMarshaller: ToResponseMarshaller[Seq[(String, String, String)]] =
+    Marshaller.withFixedContentType(ContentTypes.`text/csv(UTF-8)`) { t =>
+      val csv = CsvSerializer.asCsv(Seq("Device ID", "Failure Code", "Failure Description"), t)
+      val e = HttpEntity(ContentTypes.`text/csv(UTF-8)`, csv)
+      val h = `Content-Disposition`(ContentDispositionTypes.attachment, Map("filename" -> "device-failures.csv"))
```
jerrytrieu: Any way to put `campaignId` into filename?

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
```diff
@@ -90,8 +119,13 @@ class CampaignResource(extractAuth: Directive1[AuthedNamespaceScope],
               complete(StatusCodes.Created -> retryFailedDevices(ns, campaign, request))
             }
           } ~
-          (get & path("stats")) {
-            complete(campaigns.campaignStats(campaign.id))
+          get {
+            path("stats") {
+              complete(campaigns.campaignStats(campaign.id))
+            } ~
+            (path("failed-installations.csv") & parameter('failureCode.as[String])) {
```
jerrytrieu: Should `csv` be in the path? Could be parameter in future?

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
houcros: Updated.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
houcros: Updated, tried to make it clearer.

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
houcros: I found a way although it's not very elegant. See my last commit.

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
houcros: I remember we discussed this with @simao and we came to the conclusion that it was better in the path, although I can't recall why. I'll try to dig up that conversation. But yes, we could always change it to be in the query.

simao: IMO:

The restful way to request a format of a resource is the`Accept` header, however, many clients do not support this, for example some browsers, and it can be hard to use for some clients, so it's customary to use the file extension to request a specific format.

koshelev: We have a SPA, it should be able to manipulate headers...

simao: It wasn't last time we implemented this.

koshelev: Weird... What about adding the logic to Web App?

PR #113, opened 2019-04-02
[OTA-2479] Fix calculation of retry-campaigns status

houcros:
There is a bug in the calculation of the last `DeviceUpdate` that failed. We are finding the last update and checking if it was a failure. What we actually want is the last update that _finished_ with a failure, i.e. the last update with `status = failed` and with `status not in (requested, scheduled, accepted)`.
Otherwise if the device is e.g. in `accepted` status in a retry-campaign we won't find the last failure, because the latest status is `accepted`.

### DIFF
--- a/src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
@@ -153,12 +153,16 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
    * these campaigns, selects the most recent failures and returns them.
    */
   protected[db] def findFailedDeviceUpdatesAction(campaignIds: Set[CampaignId]): DBIO[Set[DeviceUpdate]] = {
-    Schema.deviceUpdates
-      .join(Schema.deviceUpdates
-      .filter(upd => upd.campaignId.inSet(campaignIds))
+    val latestFinishedUpdates =
+      Schema.deviceUpdates
+        .filter(_.status inSet Set(DeviceStatus.rejected, DeviceStatus.successful, DeviceStatus.cancelled, DeviceStatus.failed))
+        .filter(_.campaignId inSet campaignIds)
         .groupBy(_.deviceId)
-        .map { case (id, upd) => (id, upd.map(_.updatedAt).max) })
-        .on { (fst, snd) => fst.deviceId === snd._1 && fst.updatedAt === snd._2 }
+        .map { case (id, upd) => (id, upd.map(_.updatedAt).max) }
+
+    Schema.deviceUpdates
+      .join(latestFinishedUpdates)
+      .on { (fst, snd) => fst.deviceId === snd._1 && fst.updatedAt === snd._2 }
       .map(_._1)
       .filter(_.status === DeviceStatus.failed)
       .result
@@ -166,7 +170,7 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
   }
 
   /**
-   * Calculates campaign-wide statistic counters, also taking retry campaings
+   * Calculates campaign-wide statistic counters, also taking retry campaigns
    * into account if any exist.
    */
   def campaignStats(campaignId: CampaignId): Future[CampaignStats] = db.run {
@@ -217,7 +221,7 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
       mainCnt <- deviceUpdateRepo.countByStatus(Set(campaignId)).map(processCounts)
       retryCnt <- deviceUpdateRepo.countByStatus(retryCampaignIds).map(processCounts)
     } yield {
-      val finished = mainCnt.finished  - (retryCnt.rejected + retryCnt.cancelled)
+      val finished = mainCnt.finished - retryCnt.rejected - retryCnt.cancelled
       val failed = failedDevices.size.toLong
       val failures = processFailures(failedDevices, retryCampaigns)
 
--- a/src/test/scala/com/advancedtelematic/campaigner/db/CampaignsSpec.scala
+++ b/src/test/scala/com/advancedtelematic/campaigner/db/CampaignsSpec.scala
@@ -5,15 +5,14 @@ import cats.data.NonEmptyList
 import com.advancedtelematic.campaigner.data.DataType.CampaignStatus._
 import com.advancedtelematic.campaigner.data.DataType._
 import com.advancedtelematic.campaigner.data.Generators._
-import com.advancedtelematic.campaigner.util.CampaignerSpecUtil
-import com.advancedtelematic.campaigner.util.DatabaseUpdateSpecUtil
+import com.advancedtelematic.campaigner.util.{CampaignerSpecUtil, DatabaseUpdateSpecUtil}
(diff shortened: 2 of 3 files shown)


nkly:
Good catch! Could you also add a test for it?

houcros:
Sure, will do.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -217,7 +221,7 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
       mainCnt <- deviceUpdateRepo.countByStatus(Set(campaignId)).map(processCounts)
       retryCnt <- deviceUpdateRepo.countByStatus(retryCampaignIds).map(processCounts)
     } yield {
-      val finished = mainCnt.finished  - (retryCnt.rejected + retryCnt.cancelled)
+      val finished = mainCnt.finished - retryCnt.rejected - retryCnt.cancelled
```
houcros: I know this doesn't change the result, but it made it somehow easier for me to understand where this calculation is coming from.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -153,20 +153,24 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
    * these campaigns, selects the most recent failures and returns them.
    */
   protected[db] def findFailedDeviceUpdatesAction(campaignIds: Set[CampaignId]): DBIO[Set[DeviceUpdate]] = {
-    Schema.deviceUpdates
-      .join(Schema.deviceUpdates
-      .filter(upd => upd.campaignId.inSet(campaignIds))
+    val latestFinishedUpdates =
+      Schema.deviceUpdates
+        .filterNot(_.status inSet Set(DeviceStatus.requested, DeviceStatus.scheduled, DeviceStatus.accepted))
```
nkly: What was the reason to change it from `.filter(_.status inSet Set(DeviceStatus.successful, ...)`? I personally think it looked better the previous way.

nkly: _Side note_: we can also have `val finishedStates = Set(DeviceState.successful ...)` in `DeviceState` object as these states are used quite a lot throughout the codebase.

src/test/scala/com/advancedtelematic/campaigner/db/CampaignsSpec.scala
```diff
@@ -75,40 +74,34 @@ final class CampaignsFindFailedDevicesSpec extends AsyncFlatSpec
...
-      genResultCode = Gen.oneOf(resultCodes),
-    )).generate
+    val deviceIds = Gen.choose(2, 8).flatMap(Gen.listOfN(_, genDeviceId)).generate
+    val updates = campaignIds.flatMap(cid => deviceIds.map { did =>
+      genDeviceUpdate().map(_.copy(campaign = cid, update = updateId, device = did)).generate
+    })
 
-    val byUpdatedAtDesc = Ordering.by[DeviceUpdate, Long](_.updatedAt.getEpochSecond)
     val expectedFailures = updates
+      .filter(du => du.status != DeviceStatus.requested && du.status != DeviceStatus.scheduled && du.status != DeviceStatus.accepted)
```
nkly: Same as https://github.com/advancedtelematic/campaigner/pull/113/files#r272474463

src/test/scala/com/advancedtelematic/campaigner/db/CampaignsSpec.scala
```diff
@@ -75,40 +74,34 @@ final class CampaignsFindFailedDevicesSpec extends AsyncFlatSpec
...
+    val resultAction = for {
       _ <- Schema.campaigns ++= campaignObjects
       _ <- Schema.deviceUpdates ++= updates
       failures <- campaigns.findFailedDeviceUpdatesAction(campaignIds.toSet)
-    } yield failures).flatMap { actualFailures =>
-      actualFailures should contain theSameElementsAs expectedFailures
-    }
+    } yield failures
+
+    db.run(resultAction).flatMap(_ should contain theSameElementsAs expectedFailures)
```
nkly: _Side note_: could you please create a ticket and put a TODO to refactor all the `AsyncFlatSpec` tests? They are quite tedious to work with, IMO, and I don't see any advantages in using them instead of `ScalaFutures` mixin. If you know why `AsyncFlatSpec` might be a good deal, I'd be glad to hear it.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
houcros: It was simply for convenience :) There were less to exclude than to include. So the equivalent would be:
```.filter(_.status inSet Set(DeviceStatus.rejected, DeviceStatus.successful, DeviceStatus.cancelled, DeviceStatus.failed))```
You might be right about that being more readable. I could change it back.

houcros: About the side note: yes, that might be useful indeed. It would also give us a centralised definition of "finished". Right now it's scattered and maybe not consistent. We can refactor that in a follow-up PR.

src/test/scala/com/advancedtelematic/campaigner/db/CampaignsSpec.scala
houcros: AFIAU the main (only?) advantage is that we don't need to worry about tests failing because of a timeout when waiting for a Future to complete. I've seen a few of those and they're annoying (and the reason we need to increase the patience sometimes).
On the other hand, they _are_ tedious and make it more difficult to use property-based testing. So I guess we need to pick our poison.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
houcros: I've updated to use `filter` instead of `filterNot` to improve readability.

src/test/scala/com/advancedtelematic/campaigner/db/CampaignsSpec.scala
houcros: Updated.

src/test/scala/com/advancedtelematic/campaigner/db/CampaignsSpec.scala
nkly: I think the timeout from PatienceConfig is also applied when using `futureValue` method of `ScalaFutures` trait, so it's possible to tweak it too. At least that's the impression I get from reading http://doc.scalatest.org/3.0.1/#org.scalatest.concurrent.ScalaFutures

houcros: Indeed, but sometimes it's a bit tricky to find the right value for the patience without setting it to something ludicrous 😕 For instance in this test, I tried to use `futureValue` and it was timing out most of the times with a generous patience. 
Anyways, I'm also in favour of `ScalaFutures`, it's easier to use and to follow (and to use property-based tests). Just saying that maybe once in a while `AsyncFlatSpec` is worth the trouble.

PR #114, opened 2019-04-04
Fix launching retry-campaign when a device fails with a new code

houcros:
Currently, if a device fails in a retry-campaign with a failure code that is different than the one it originally failed with, we are not able to take it into account when launching a new retry-campaign for the new failure code. Instead we get a `MissingFailedDevices` error because we're only searching for devices that failed under the main campaign. This commit fixes that and adds a test for it.

### DIFF
--- a/src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
@@ -107,22 +107,14 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
   }
 
   /**
-    * Returns the IDs of all the devices that were processed in the campaign with `campaignId` and failed with
-    * the code `failureCode`.
+    * Returns the most recent device updates that have failed with the code `failureCode` in the campaign with ID
+    * `mainCampaignId` or any of its retry-campaigns.
     */
-  def fetchFailedDevices(campaignId: CampaignId, failureCode: String): Future[Set[DeviceId]] =
-    deviceUpdateRepo.findFailedByFailureCode(campaignId, failureCode)
-
-  /**
-    * Returns all the devices that are now failed with the code `failureCode` as a result of running the campaign
-    * with ID `campaignId` or any of its retry-campaigns.
-    */
-  def fetchFailureCodes(campaignId: CampaignId, failureCode: String): Future[Set[(DeviceId, String, String)]] = db.run {
+  def findLatestFailedUpdates(mainCampaignId: CampaignId, failureCode: String): Future[Set[DeviceUpdate]] = db.run {
     campaignRepo
-      .findRetryCampaignsOfAction(campaignId)
-      .flatMap(cids => findFailedDeviceUpdatesAction(cids.map(_.id) + campaignId))
+      .findRetryCampaignsOfAction(mainCampaignId)
+      .flatMap(cids => findFailedDeviceUpdatesAction(cids.map(_.id) + mainCampaignId))
       .map(_.filter(_.resultCode.contains(failureCode)))
-      .map(_.map(du => (du.device, du.resultCode.getOrElse(""), du.resultDescription.getOrElse(""))))
   }
 
   def countByStatus: Future[Map[CampaignStatus, Int]] =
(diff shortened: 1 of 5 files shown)


src/test/scala/com/advancedtelematic/campaigner/http/CampaignResourceSpec.scala
```diff
@@ -472,4 +485,25 @@ class CampaignResourceSpec
     val elemsWithGroupNumAssigned = seq.groupBy(_ => Gen.choose(0, n - 1).generate)
     0.until(n).map(groupNum => elemsWithGroupNumAssigned.getOrElse(groupNum, Seq.empty))
   }
+
+  private def assertRetryCampaign(retryCampaign: GetCampaign, failureCode: String, updateId: UpdateId, mainCampaignId: CampaignId): Assertion = {
+    eventually {
```
nkly: What was the reason to wrap it in `eventually`? Shouldn't the retry campaign be available right after we get 201 from the `retry-failed` endpoint?

houcros: The thing is that sometimes the test checks too soon and the status of the retry-campaign is `prepared` instead of `launched`, making the test flaky. My two ideas would be to wrap it in an `eventually` or to leave the campaign status out of the comparison. It just seemed easier to go with `eventually`.

nkly: Ah, I see, good point

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -107,21 +107,30 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
-    * with ID `campaignId` or any of its retry-campaigns.
+    * Returns the deviceIds of the most recent device updates that have failed with the code `failureCode`
+    * in the campaign with ID `mainCampaignId` or any of its retry-campaigns.
     */
-  def fetchFailureCodes(campaignId: CampaignId, failureCode: String): Future[Set[(DeviceId, String, String)]] = db.run {
-    campaignRepo
-      .findRetryCampaignsOfAction(campaignId)
-      .flatMap(cids => findFailedDeviceUpdatesAction(cids.map(_.id) + campaignId))
-      .map(_.filter(_.resultCode.contains(failureCode)))
+  def fetchLatestFailedDevices(campaignId: CampaignId, failureCode: String): Future[Set[DeviceId]] = db.run {
```
koshelev: I would suggest we create a generic `def fetchFailedDevices(campaignId: CampaignId, failureCode: String): Future[Set[DeviceUpdate]]` in `Campaigns` class and left it to caller to extract attributes it needs. Each of these two method has exactly one usage and I think the knowledge of what exactly caller needs should be a part of `Campaigns`

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -107,21 +107,30 @@ protected [db] class Campaigns(implicit db: Database, ec: ExecutionContext)
...
+    findLatestFailedUpdates(campaignId, failureCode)
+      .map(_.map(_.device))
+  }
+
+  /**
+    * For all the most recent device updates that have failed with the code `failureCode` in the campaign with ID
+    * `mainCampaignId` or any of its retry-campaigns, return the deviceId, failure code and failure description.
+    */
+  def fetchLatestFailureCodes(campaignId: CampaignId, failureCode: String): Future[Set[(DeviceId, String, String)]] = db.run {
```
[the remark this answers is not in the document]
houcros: Indeed, but this info comes originally from the installation reports. So I have a PR open to use a `ResultCode` value class when generating the reports: https://github.com/advancedtelematic/libats/pull/100. I could add a `ResultDescription` there as well. Then from the reports we can propagate it anywhere used.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
houcros: Makes sense. Updated.

src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
```diff
@@ -62,11 +62,13 @@ class CampaignResource(extractAuth: Directive1[AuthedNamespaceScope],
...
-      Future.traverse(_) { case (did, fc, fd) =>
-        deviceRegistry.fetchOemId(ns, did).map((_, fc, fd))
-      }
-      .map(_.toSeq)
+    val f = campaigns
+      .findLatestFailedUpdates(campaignId, failureCode)
+      .map(_.map(du => (du.device, du.resultCode.getOrElse(""), du.resultDescription.getOrElse(""))))
+      .flatMap {
+        Future.traverse(_) { case (did, fc, fd) =>
+          deviceRegistry.fetchOemId(ns, did).map((_, fc, fd))
```
koshelev: I can imagine following problems with fetching ids individually:

- a lot of requests to device registry
- request times out because it takes too long to complete the future

houcros: Indeed, we might run into that kind of issues. I proposed a new endpoint in device-registry to fetch all this IDs at once but it wasn't well received. I can create a new ticket to discuss a different approach to improve this. However, this wasn't the bug I was trying to fix in this PR.

houcros: @koshelev I don't think this ought to be a blocker for this bugfix. Do you think I can merge this?

PR #116, opened 2019-04-26
Stream and limit requested devices from the DB

nkly:
Fixes OTA-2385. No pressure on reviewing it, but it would be nice to merge by the end of the year :)

### DIFF
--- a/src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
@@ -1,6 +1,7 @@
 package com.advancedtelematic.campaigner.actor
 
 import akka.actor.{Actor, ActorLogging, Props, Status}
+import akka.stream.ActorMaterializer
 import com.advancedtelematic.campaigner.client._
 import com.advancedtelematic.campaigner.data.DataType._
 import com.advancedtelematic.campaigner.db.{Campaigns, DeviceUpdateProcess}
@@ -41,6 +42,8 @@ class CampaignScheduler(director: DirectorClient,
   private val campaigns = Campaigns()
   private val deviceUpdateProcess = new DeviceUpdateProcess(director)
 
+  implicit val materializer = ActorMaterializer.create(context)
+
   override def preStart(): Unit =
     self ! NextBatch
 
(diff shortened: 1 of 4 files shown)


src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
```diff
@@ -54,9 +57,10 @@ class CampaignScheduler(director: DirectorClient,
   def receive: Receive = {
     case NextBatch =>
       log.debug("Requesting next batch")
-      // TODO (OTA-2383) limit or stream device IDs from DB
-      campaigns.requestedDevices(campaign.id)
-        .map(deviceIds => BatchToSchedule(deviceIds.take(batchSize)))
+      campaigns.requestedDevicesStream(campaign.id)
+        .take(batchSize.toLong)
+        .runFold(Set.empty[DeviceId])(_ + _)
+        .map(BatchToSchedule(_))
```
houcros: `.map(BatchToSchedule)`

src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
```diff
@@ -54,9 +57,10 @@ class CampaignScheduler(director: DirectorClient,
   def receive: Receive = {
     case NextBatch =>
       log.debug("Requesting next batch")
-      // TODO (OTA-2383) limit or stream device IDs from DB
-      campaigns.requestedDevices(campaign.id)
-        .map(deviceIds => BatchToSchedule(deviceIds.take(batchSize)))
+      campaigns.requestedDevicesStream(campaign.id)
```
koshelev: Is not some special statement/connection configuration required to configure a page size jdbc client should use? I can imagine default behaviour will be for driver fetch all the data and than Slick will turn it into a stream

nkly: Hmm, though there is no mention of it in Akka Streams docs, it seems like you're right: https://mariadb.com/kb/en/library/about-mariadb-connector-j/#streaming-result-sets. Will fix it.

nkly:
@koshelev Ah, nope, it's just a little cleanup of the things left undone after "Retry update" feature implementation.

PR #117, opened 2019-05-02
Bump libats version

houcros:
Necessary after https://github.com/advancedtelematic/libats/pull/100.
Use libats `ResultCode` and `ResultDescription` value classes for installation results.


houcros:
Yes, my bad; a couple of those cosmetic changes snuck in and I didn't bother to revert them. I'd be happy to use `scalafmt`. I think it is/was used in some services but generally we ignore it. We could start with campaigner and move on from there.

houcros:
I wasn't able to find a canonical `scalafmt` configuration that conforms to the [Scala Style Guide](https://docs.scala-lang.org/style/), assuming it exists. There was some talk about creating one [here](https://contributors.scala-lang.org/t/creation-of-a-standard-scala-style-guide/933) but it seems it didn't catch on. We could create our own formatting file according to the style guide to avoid holy wars. I can start working on that.

PR #118, opened 2019-05-08
Do not handle device update events with MTU correlation ID

nkly:
MTU are not related to campaigns, and thus should not be used to change
the state of campaign devices, especially not in _all_ the campaigns.
According to @jerrytrieu, this "overly backward compatible" thing was
implemented to because
 
> Before correlation ids devices in campaigns had mtuid only, So some devices may complete a campaign long after it started. This was to handle that case

and should be safe to remove now.


nkly:
Ah, stupid me just merged another refactor into the branch. Please ignore the 2nd commit and check only the 1st 'cause it's the one that's relevant to the PR.

PR #120, opened 2019-05-13
[OTA-2356] Communicate with Director asynchronously

nkly:


nkly:
Knock-knock @koshelev @jerrytrieu. It would be nice if you could check it out.

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
```diff
@@ -176,9 +176,9 @@ object DataType {
    *   to `processed` state in the UI until a device goes to the Director)
    * - `rejected` when a device is rejected by Director and is not a part of the
    *    campaign anymore (corresponds to `not impacted` state in the UI)
-   * - `scheduled` when a device is approved by Director and is scheduled for
-   *    an update (partly corresponds to `queued` state in the UI)
-   * - `accepted` when an update was accepted on a device and is about to be
+   * - `scheduled` when a device is scheduled for an approval by user and
+   *    Director (partly corresponds to `queued` state in the UI)
```
jerrytrieu: Should just be `device has been assigned an update by Director`. There's no user input at this state.

jerrytrieu:
Let's put this on berlin hat and run e2e.

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
nkly: What you've described actually corresponds to `accepted` state. `scheduled` is the state we're in after sending an assignment request to director, but before receiving approval/rejection response from it.

build.sbt
```diff
@@ -25,7 +25,7 @@ resolvers += "ATS Snapshots" at "http://nexus.advancedtelematic.com:8081/content
 libraryDependencies ++= {
   val akkaV = "2.5.13"
   val akkaHttpV = "10.1.1"
-  val libatsV = "0.2.1-13-g4a34692"
+  val libatsV = "0.2.1-18-g7726a72"
```
koshelev: Does the version of `libats` exists?

nkly: Nope, it's the one I published locally with changes from advancedtelematic/libats#95 because that PR is not merged yet.

nkly: I think the original idea of @jerrytrieu was to figure out whether it is possible to switch campaigner <-> director comms to messages and then merge advancedtelematic/libats#95 when it's done.

PR #124, opened 2019-05-29
Add scalafmt and linters

nkly:
Let's discuss code formatting. I believe that our status quo where many different styles live together, sometimes within the same file, is tiresome to read and write. I'd be very glad to see the code _consistent_, regardless of which particular style is used. I also acknowledge that we're a team and we need to find a common ground on this issue.

I think it's a good start to add an auto-formatter to Campaigner, _gradually_ re-format the code (i.e. when you touch a file, you suppose to re-format it) and keep it well-formatted via CI check. After a while we can evaluate whether this approach brings us any value and/or causes pain and maybe extend it to other services or cease completely. 

Feel free to share your ideas and thoughts.

----

Other notes:

- The scalafmt settings were chosen to be as much close to the Scala Style guide as possible ([link](https://contributors.scala-lang.org/t/creation-of-a-standard-scala-style-guide/933/12))


simao:
My opinion is against using these tools.

I think code style is just that, a style, and everyone has their own style, not sure what improvement enforcing a style on everyone brings. 

These are are an extra barrier on a developer and finishing a task and they also add more dependencies to an already complex build.

I also don't understand how code style makes code "tiresome to read and write". To me, having a tool change half the file while I am trying to change that file is more tiresome than not having a code style. Huge diffs because a tool changed a whole file are also not very helpful.

nkly:
Thanks @simao for sharing your opinion. Before proceeding with discussion I'd like to find out whether others are interested in this topic and what is their stance on it. @koshelev @houcros @jerrytrieu I'd be glad if you could share your opinions as well or explicitly waive from the discussion if you have no interest in it.

houcros:
It's a delicate topic.

On the one hand I think we can benefit from it, in the sense that it would be easier to switch context and work in different repos. I think that's what was behind Nikolai's "tiresome to read" point. I think this should also help with the diffs eventually, e.g. we wouldn't check-in reordering of imports.

On the other hand, sometimes it's gonna force us to read code in a way that we're not used to or don't like, because we'll never agree on a style and we'll have to compromise.

All in all, I actually don't have a strong opinion about this topic. I think it wouldn't affect me much if we either keep going as until now or introduce some (reasonable) formatting.

PR #128, opened 2019-07-24
[OTA-2661] Do not update the status of a campaign if it was cancelled

nkly:
Previously the status of a campaign was updated each time a new device
finishes. A campaign could get `cancelled` status when it had all its
groups cancelled. When switching from using groups to sets of
devices, this functionality was lost and a campaign could never get
`cancelled` status based on statuses of its devices.

This PR disables updating campaign's status when it's set as `cancelled`
by the code in `/api/v2/campaigns/<uuid>/cancel` endpoint.

NOTE: the status won't be changed to `failed` even if some cancelled
devices fail. Once it's cancelled, it won't EVER be changed. If this 
behaviour is not desirable, e.g. the campaign should get `failed` status
even though it was `cancelled` earlier, it could be fixed.

### DIFF
--- a/src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
@@ -4,7 +4,7 @@ import java.time.Instant
 
 import akka.NotUsed
 import akka.stream.scaladsl.Source
-import cats.syntax.either._
+import cats.syntax.option._
 import com.advancedtelematic.campaigner.data.DataType.CampaignStatus.CampaignStatus
 import com.advancedtelematic.campaigner.data.DataType.DeviceStatus.DeviceStatus
 import com.advancedtelematic.campaigner.data.DataType.SortBy.SortBy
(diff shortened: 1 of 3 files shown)


src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -327,7 +328,8 @@ protected [db] class CampaignStatusTransition(implicit db: Database, ec: Executi
       }
     } yield ()
 
-  protected [db] def calculateCampaignStatus(campaign: CampaignId): DBIO[Either[Unit, CampaignStatus]] = {
+  protected [db] def calculateCampaignStatus(
+      campaign: CampaignId, currentCampaignStatus: Option[CampaignStatus] = None): DBIO[Either[Unit, CampaignStatus]] = {
```
houcros: Having `DBIO[Either[Unit, CampaignStatus]]` as return type feels odd... Could this be `DBIO[Option[CampaignStatus]]`?

nkly: I guess so.

PR #129, opened 2019-07-25
[OTA-2450] Device can cancel update from a campaign pending approval

houcros:
Campaigner should only offer an endpoint to cancel updates for a device in a campaign that is pending approval, since these are not known to director yet, i.e. there is no assignment in director.

For the other cases, when an update has been assigned in director (a campaign update or a MTU) we should call director at `DELETE /api/v1/assignments/:deviceUuid` to cancel the update. Director will publish a `DeviceUpdateCancelled` event and, if the update came from a campaign, campaigner process the event and cancel the update.

**Note:** this comes with some changes in the FE to use the corresponding endpoints: https://github.com/advancedtelematic/ota-plus-server/pull/2034

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.2.7
+sbt.version=1.2.8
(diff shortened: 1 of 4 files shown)


houcros:
I added one more commit to make this precondition explicit, i.e. this endpoint should fail when trying to cancel an update from a campaign that is approved automatically. In that case we should use the director 'assignments' resource.

PR #131, opened 2019-08-01
Added exclusion of test scope

jochenschneider:
Signed-off-by: Jochen Schneider <j.schneider@here.com>


houcros:
Is this to appease the OSS scanner? How does this work actually? Sorry haven't seen such a thing before (except your previous PR).

PR #134, opened 2019-08-02
[OTA-3153] Re-calculate the status of a campaign when all devices are scheduled

nkly:
It prevents an edge-case situation when a campaign for an empty group gets stuck in `launched` state and never finishes.

### DIFF
--- a/src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
@@ -10,6 +10,7 @@ import slick.jdbc.MySQLProfile.api._
 
 import scala.concurrent.Future
 import scala.concurrent.duration._
+import scala.util.Success
 
 object CampaignScheduler {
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
```diff
@@ -68,9 +68,11 @@ class CampaignScheduler(director: DirectorClient,
       schedule(devices).pipeTo(self)
 
     case BatchToSchedule(devices) if devices.isEmpty =>
-      parent ! CampaignComplete(campaign.id)
-      // TODO: Should move to finished
-      context.stop(self)
+      campaigns.updateStatus(campaign.id).onComplete { _ =>
```
houcros: Why not `forEach` if we don't handle errors here?

src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
```diff
@@ -68,9 +68,11 @@ class CampaignScheduler(director: DirectorClient,
...
+        log.debug(s"Completed campaign: ${campaign.id}")
```
houcros: Isn't this logged in `CampaignSupervisor` already?

src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
nkly: To make sure that the actor stops no matter what happens in `updateStatus`

src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
nkly: True. Guess it's just a leftover statement I used to debug tests. I believe it won't hurt much if leave it there :)

src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
houcros: Yeah that make sense to me 👍

houcros:
Could we lose the merge commit though?

nkly:
@houcros Sure I'll squash it before merging

src/main/scala/com/advancedtelematic/campaigner/actor/CampaignScheduler.scala
simao: It's not safe to call `stop` on `onComplete`. The actor might be processing another message by the time `onComplete` is executed. See 2nd warning on this section of the akka docs: https://doc.akka.io/docs/akka/current/actors.html#ask-send-and-receive-future

nkly: Thanks @simao, didn't know that. I'll update the code to avoid calling `stop` in the callback

PR #135, opened 2019-08-26
[OTA-3153] Disallow creating campaigns without devices

houcros:
Campaigns without devices (more precisely, with device groups without devices) are not finishing. This is because the campaign supervisor is simply not picking them up, since the logic to select the new campaigns only looks at the devices.
However, this use case shouldn't be allowed. So the changes in this PR are to disallow creating campaigns without devices (an error will be thrown if attempted).

### DIFF
--- a/src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/http/CampaignResource.scala
@@ -37,10 +37,12 @@ class CampaignResource(extractAuth: Directive1[AuthedNamespaceScope],
   def createCampaign(ns: Namespace, request: CreateCampaign): Future[CampaignId] = {
     val campaign = request.mkCampaign(ns)
     val metadata = request.mkCampaignMetadata(campaign.id)
-    for {
-      devices <- fetchDevicesInGroups(ns, request.groups)
-      campaignId <- campaigns.create(campaign, request.groups.toList.toSet, devices, metadata)
-    } yield campaignId
+    fetchDevicesInGroups(ns, request.groups).flatMap {
+      case devices if devices.isEmpty =>
+        Future.failed(Errors.CampaignWithoutDevices)
+      case devices =>
+        campaigns.create(campaign, request.groups.toList.toSet, devices, metadata)
+    }
   }
 
   def searchCampaigns(ns: Namespace): Route =
--- a/src/main/scala/com/advancedtelematic/campaigner/http/Errors.scala
+++ b/src/main/scala/com/advancedtelematic/campaigner/http/Errors.scala
@@ -20,6 +20,7 @@ object ErrorCodes {
   val ConflictingUpdate = ErrorCode("update_already_exists")
   val TimeoutFetchingUpdates = ErrorCode("timeout_fetching_updates")
   val CampaignDoesNotRequireApproval = ErrorCode("campaign_does_not_require_approval")
+  val CampaignWithoutDevices = ErrorCode("campaign_without_devices")
 }
 
 object Errors {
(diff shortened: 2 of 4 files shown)


project/Release.scala
```diff
@@ -1,13 +1,7 @@
...
-import com.typesafe.sbt.packager.SettingsHelper._
-import sbtrelease._
-import sbtrelease.ReleaseStateTransformations.{setReleaseVersion => _, _}
-import sbt.Keys._
-import sbt._
 import com.typesafe.sbt.SbtNativePackager.Docker
-
+import sbt.Keys._
 import sbtrelease.ReleasePlugin.autoImport._
+import sbtrelease.ReleaseStateTransformations.{setReleaseVersion => _, _}
```
koshelev: why do we need it?

houcros:
> Why are we allowing to create campaigns without devices in first place?

Fair point. Do we want to fail in that case?
OTA-3153 is about "campaigns without devices don't finish" though. If we don't want to allow campaigns without devices I will close that and create a new bug about "campaigns without devices are allowed". (I wish we had groomed that better.)

project/Release.scala
houcros: Right, we only need `checkSnapshotDependencies` actually. Updated.

houcros:
Updated: not allowing to create campaigns without devices.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -336,10 +336,8 @@ protected [db] class CampaignStatusTransition(implicit db: Database, ec: Executi
     for {
       affected <- devicesWithStatus(DeviceStatus.values - DeviceStatus.requested - DeviceStatus.rejected)
       finished <- devicesWithStatus(Set(DeviceStatus.successful, DeviceStatus.failed, DeviceStatus.cancelled))
-      cancelled <- devicesWithStatus(Set(DeviceStatus.cancelled))
       requested <- devicesWithStatus(Set(DeviceStatus.requested))
-      total <- devicesWithStatus(DeviceStatus.values)
-      wasCampaignCancelled = currentCampaignStatus.exists(_ == CampaignStatus.cancelled)
+      wasCampaignCancelled = currentCampaignStatus.contains(CampaignStatus.cancelled)
```
koshelev: The whole method is kind of weird. From what i see it should be `isFinished(...): Boolean`

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
```diff
@@ -336,10 +336,8 @@ protected [db] class CampaignStatusTransition(implicit db: Database, ec: Executi
     for {
       affected <- devicesWithStatus(DeviceStatus.values - DeviceStatus.requested - DeviceStatus.rejected)
       finished <- devicesWithStatus(Set(DeviceStatus.successful, DeviceStatus.failed, DeviceStatus.cancelled))
```
koshelev: Is there a reason we are not using something like `select count(*), status from device_updates where campaign_id=? group by status`?

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
houcros: Indeed. Refactored this in a new PR at #138 since this doesn't seem so minor anymore.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
houcros: Ditto.

src/main/scala/com/advancedtelematic/campaigner/db/Campaigns.scala
koshelev: should we close this one and move to the new one?

houcros: Yes, just that this one is rebased on #138 now. So only the second commit of this branch should be reviewed here.

houcros:
Reviewed in person with @koshelev.

PR #137, opened 2019-08-27
Fix flaky test.

houcros:


src/test/scala/com/advancedtelematic/campaigner/http/CampaignResourceSpec.scala
```diff
@@ -534,7 +536,7 @@ class CampaignResourceSpec
         Nil,
         autoAccept = true
       )
-      retry.createdAt shouldBe retry.updatedAt
+      retry.createdAt.truncatedTo(ChronoUnit.MINUTES) shouldBe retry.updatedAt.truncatedTo(ChronoUnit.MINUTES)
```
jochenschneider: Would it maybe make sense to test that they are not too far apart?

houcros: Yes, I think that's more reasonable actually. Updated.

PR #140, opened 2019-09-11
Add gitlab CD integration for static code scans

zabbal:
The veracode pipeline is intended to be run as part of schedule only.


zabbal:
https://main.gitlab.in.here.com/olp/edge/ota/connect/back-end/campaigner/pipelines/140364 - successful test run.

PR #141, opened 2019-09-11
[OTA-3668] Add query parameter to fetch campaigns with errors.

houcros:
Needed to show failed campaigns in the dashboard of the new UI.


houcros:
Anyone against me merging this?

src/main/scala/com/advancedtelematic/campaigner/data/DataType.scala
```diff
@@ -121,6 +122,20 @@ object DataType {
...
   final case class UpdateCampaign(name: String, metadata: Option[Seq[CreateCampaignMetadata]] = None)
 
+  final case class SearchCampaignParams(status: Option[CampaignStatus],
+                                        nameContains: Option[String],
+                                        withErrors: Option[Boolean],
+                                        sortBy: SortBy,
+                                        offset: Long,
+                                        limit: Long,
+                                       ) {
+    if (withErrors.isDefined) {
```
koshelev: Are we handling `IllegalArgumentException` somewhere?

houcros: Yes, Akka Http takes care of it automatically for this "case class extraction" and returns a 400. See https://doc.akka.io/docs/akka-http/current/routing-dsl/case-class-extraction.html#case-class-validation.

PR #142, opened 2019-09-19
Bump libtuf, libats, add migrations main class

simao:


simao:
cc @SusVlaAlex You'll need this

PR #147, opened 2019-11-14
Support result descriptions bigger than 255 Chars

simao:
Increase the database column that saves the description to 1024 chars.

Cap incoming result descriptions that are bigger than 1024 chars and
save only the first part.

This assumes both the database and the jvm are using UTF-8 to transfer
the result description.

Signed-off-by: Simão Mata <simao.mata@here.com>


rdanitz:
Hmm are gitlab pipeline tests known to be flaky?

simao:
Yeah :\

